PR #8, opened 2024-01-12
fix: remote create new file

BruceMacD:
Just noticed this while working on the `ollama-js` library, two bugs.
- A new-line in the Modelfile causes `path.exists()` to not be true.
- The blob creation endpoint is `POST` rather than `PUT`

Example Modelfile:
```
FROM ~/models/nous-capybara/nous-capybara-34b.Q4_0.gguf
TEMPLATE "USER: { .Prompt } ASSISTANT: "
```


jmorganca (approved):

mxyng (approved):
Good catch!

PR #16, opened 2024-01-19
fix: update OLLAMA_HOST parsing to match ollama CLI

mxyng:


jmorganca (approved):

ollama/_client.py
```diff
@@ -607,3 +605,52 @@ def _as_bytesio(s: Any) -> Union[io.BytesIO, None]:
   elif isinstance(s, bytes):
     return io.BytesIO(s)
   return None
+
+
+def _parse_host(host: str) -> str:
+  """
+  >>> _parse_host('')
```
mxyng, line 612: these are the same(ish) test cases as those in `ollama/api/client_test.go`

PR #17, opened 2024-01-22
update response error field to match json response

mxyng:
also formatting because I didn't run format earlier

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -11,31 +11,31 @@ else:
 
 class BaseGenerateResponse(TypedDict):
   model: str
-  "Model used to generate response."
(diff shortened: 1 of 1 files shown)


BruceMacD (approved):
lgtm

ollama/_types.py
```diff
@@ -137,25 +137,25 @@ class RequestError(Exception):
   def __init__(self, content: str):
     super().__init__(content)
     self.content = content
-    "Reason for the error."
+    'Reason for the error.'
```
BruceMacD, line 140: Should `RequestError` be `error` rather than `content` also?

PR #61, opened 2024-02-12
Fixes URL parsing when basic auth is included in the URL.

lemeur:
Ollama may be protected by a reverse proxy enforcing basic auth.

When the Ollama URL contains basic auth elements, ollama-python removed them from the URL leading in a HTTP 401 ERROR.
```
from ollama import Client
client = Client(host='http://<username>:<password>@<ip>:<port>/')
```

This patch fixes the parsing of the URL so that basic auth elements are not removed.

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -718,8 +718,10 @@ def _parse_host(host: Optional[str]) -> str:
   elif scheme == 'https':
     port = 443
 
-  split = urllib.parse.urlsplit('://'.join([scheme, hostport]))
+  split = urllib.parse.urlsplit('://'.join([scheme, hostport]), allow_fragments=True)
   host = split.hostname or '127.0.0.1'
   port = split.port or port
-
-  return f'{scheme}://{host}:{port}'
+  parsed_url = f'{scheme}://{host}:{port}'
+  if split.username and split.password:
+    parsed_url=f'{scheme}://{split.username}:{split.password}@{host}:{port}'
+  return parsed_url


ChaseMB00:
Ran into this Problem today. Would be really nice if this pull request would be accepted

TheMcSebi:
I also ran into this issue quite a while ago and I can also confirm that this solves the issue.

TheMcSebi:
I would really love to see this fix on main

imhimansu28 (approved):

Padi2312:
Any updates here ?

TheMcSebi:
No updates? I'm still relying on installing a custom wheel in internal deployments

ParthSareen:
Thanks for echoing your thoughts here everyone. Going to give this some thought and come back to ya'll!

manykarim:
Just wanted to mention:
You can use the following workaround to use HTTP Auth with the Ollama Python package:

```
from ollama import Client
import httpx
httpx_auth = httpx.BasicAuth(username="myusername", password="mypassword")
client = Client(host='https://ollama.service:32559', auth=httpx_auth)
response = client.chat(model='llama3.1', messages=[
  {
    'role': 'user',
  }])
print(response)
```

Found it here https://medium.com/@yuxiaojian/host-your-own-ollama-service-in-a-cloud-kubernetes-k8s-cluster-c818ca84a055

Tried it and it works

TheMcSebi:
I'll definitely try this, it doesn't even look like a workaround but instead like the preferred method. I wonder why I didn't find this in the docs.

shivams:
Yep. And if you're using Bearer Token Authentication, then you can just pass the token in headers. 

```python
from ollama import Client

headers = {
    "Authorization": "Bearer MySecureToken"
}

client = Client(host='http://myserver.mydomain', headers=headers)
response = client.chat(
    model='llama3.3', 
    messages=[
        {
            'role': 'user',
            'content': 'Who art thou?',
        }
    ]
)
print(response)
```

josx:
I came across with this problem. This is a base dep for every software trying to use ollama, so the result is every app using this client need to rewrite the same feature.

Really it is a core feature.

PR #42, opened 2024-01-30
bump httpx to 0.26.0

ej52:
I am the creator on [Ollama Conversation](https://github.com/ej52/hass-ollama-conversation), I am wanting to get the integration listed on Home Assistant itself but in order to do that this condition needs to be met [communication-with-devicesservices](https://developers.home-assistant.io/docs/creating_component_code_review#4-communication-with-devicesservices).

So I need to switch from making direct API calls to using this library, however the version of **httpx** required by this library has conflicts with Home Assistant requirements.

This PR just bumps httpx version to 0.26.0 in order to meet the requirements.

PR #97, opened 2024-03-22
Loosen httpx constraint

balloob:
This loosens the httpx constraint to allow any version <1. I have pinned httpx as a dev dependency to the latest version of httpx (0.27.0) so tests don't just start breaking one day.

Pinning dependencies in a library forces downstream consumers to follow the requirements of the library. In the case of Home Assistant, this is not compatible as Home Assistant aggressively follows the latest version of httpx.

Unblocks https://github.com/home-assistant/core/pull/113962

Replaces #42

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/jmorganca/ollama-python"
 
 [tool.poetry.dependencies]
 python = "^3.8"
-httpx = "^0.25.2"
+httpx = "<1"
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^7.4.3"
@@ -19,6 +19,7 @@ pytest-cov = "^4.1.0"
 pytest-httpserver = "^1.0.8"
 pillow = "^10.2.0"
 ruff = "^0.1.8"
+httpx = "0.27.0"
 
 [build-system]
 requires = ["poetry-core"]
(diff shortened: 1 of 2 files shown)


balloob:
@mxyng sorry for the ping. We're preparing a new Home Assistant release this Wednesday and would like to include an Ollama integration (https://github.com/home-assistant/core/pull/113962), but it depends on this PR. Would you have time today or tomorrow to take a look? Thanks!

pyproject.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/jmorganca/ollama-python"
 
 [tool.poetry.dependencies]
 python = "^3.8"
-httpx = "^0.25.2"
+httpx = "<1"
```
BruceMacD, line 13:
```suggestion
httpx = ">=0.25.2,<1"
```

Thoughts on this to preserve the current compatibility?

synesthesiam: That should work fine since Home Assistant is on `httpx==0.27.0` :+1:

mxyng:
Setting this kind of constraint on a project pre-1.0 is risky since semantic versions are undefined at that stage. It's entirely possible (though unlikely) httpx==0.28.0 includes a breaking change. It would be better to keep httpx (and any other runtime dependencies) up to date than to have such a broad version requirement

balloob:
The risk can be avoided by the user to pin httpx to the version that they want and/or need for their application. This shouldn't be determined by one of their dependencies. What if another dependency had another opinion! 

For anyone interested in this feature, we are [maintaining a fork](https://pypi.org/project/ollama-hass/) with the loosened httpx constraint.

PR #118, opened 2024-04-15
fix 'content' check on messages in chat()

DSLstandard:
In `chat()` in `Client` and `AsyncClient`, if `message['content']` is `""`,

the following would raise an error since `bool("")` also returns `False` like `bool(None)`:
```python
if not message.get('content'):
  raise RequestError('messages must contain content')
```
Doing the following would fix it:
```python
if 'content' not in message:
  raise RequestError('messages must contain content')
```

As far as I know ollama does not explicitly prohibit empty content in messages.


mxyng (approved):

mxyng:
Thanks!

PR #130, opened 2024-04-25
Updated readme to mention that this library needs ollama service to be running and also snippet to run it if someone is using a jupyter notebook

omkar787:
Note added in readme about the fact that the ollama service should be running before using ollama python library and also details about how to solve the connection refused error while running ollama in a jupyter notebook is also added

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,23 @@ response = ollama.chat(model='llama2', messages=[
 print(response['message']['content'])
 ```
 
+### Note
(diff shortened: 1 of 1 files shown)


WolfTheDeveloper (approved):
Yes, this will be very helpful for newcomers

omkar787:
> Yes, this will be very helpful for newcomers

Yes 🚀

ParthSareen:
Hey! Thanks for this. I think to keep the pattern simple users should just be running Ollama on a separate terminal. The new docs clarify some of this!

PR #132, opened 2024-04-26
Add type overloads for chat() method in _client.py

simonottenhauskenbun:
This PR adds two type overloads in `_client.py` to reduce type warnings.
The PR should only affect typing and no logic.

## Before PR:

```python
import ollama
response = ollama.chat(model='llama2', messages=[
  {
    'role': 'user',
  },
])
# before PR: response is now Union[Mapping[str, Any], Iterator[Mapping[str, Any]]]
print(response['message']['content'])
```

## New behaviour:

```python
response = ollama.chat(model='llama2', stream=False, ...)
# response is of type Mapping[str, Any]

response = ollama.chat(model='llama2', stream=True, ...)
# response is of type Iterator[Mapping[str, Any]]

response = ollama.chat(model='llama2', ...)
# response is also of type Mapping[str, Any], since stream=False is the default.
```

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -10,7 +10,7 @@ from pathlib import Path
 from hashlib import sha256
 from base64 import b64encode, b64decode
 
-from typing import Any, AnyStr, Union, Optional, Sequence, Mapping, Literal
+from typing import Any, AnyStr, Union, Optional, Sequence, Mapping, Literal, overload
 
 import sys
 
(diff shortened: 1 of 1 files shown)


mxyng:
This is a good change. I was hesitant to add it originally since there are so many methods that should be overloaded.

@royjhan is going to carry this change if it's ok with you @simonottenhauskenbun and implement overloads for the other streaming methods

simonottenhauskenbun:
> This is a good change. I was hesitant to add it originally since there are so many methods that should be overloaded.
> 
> @royjhan is going to carry this change if it's ok with you @simonottenhauskenbun and implement overloads for the other streaming methods

Sorry for the late replay, I was on vaccation :)
Thanks for implementing all the overloads. This PR is now longer needed.

PR #141, opened 2024-05-01
Added temperature parameter

Codyk12:
Added parameter to include temperature


WolfTheDeveloper (approved):
Looks Good

mxyng:
Temperature is already configurable through [options](https://github.com/ollama/ollama-python/blob/main/ollama/_types.py#L121) so this is redundant.

PR #114, opened 2024-04-05
modified embeddings return annotation

N-Vlahovic:
Modified the annotation of the return statement in the `embeddings` endpoint from `Sequence[float]` to `Mapping[str, Sequence[float]]`.

Reasoning:
- Ollama's embeddings endpoint currently returns the following structure `{"embedding": [...]}`
- Example:
```
curl http://localhost:11434/api/embeddings -d '{
  "model": "nomic-embed-text",
  "prompt": "The sky is blue because of Rayleigh scattering"
}'


[...]

PR #145, opened 2024-05-03
fix: annotation in AsyncClient.embedding

jingfelix:
The same question as #114 .

In #114, annotation of `Client.embedding` is modified, but `AsyncClient` is forgotten.


mxyng:
Oops thanks for fixing!

PR #170, opened 2024-05-29
Add URL path to client URL in in Client._parse_host()

### LINKED ISSUE #113: Client fails to query resources behind reverse proxy path
The `_parse_host` method seems to be stripping off the end of URLs that Ollama may be proxied behind. For example, if I have a rule setup in Caddy/Nginx ro forward Ollama to http://localhost:8080/ollama, the client returns "http://localhost:8080", which causes all API requests to fail as the URL is technically invalid.

### LINKED ISSUE #244: Client's host doesn't work when the host is behind a proxy server
I hosted a Ollama server on a K8s using Helm Charts and I add a proxy server(APISix) for this deployment. Therefore, I can access the Ollama service via a host with path instead of using port-forwording.

For example, the url for the Ollama service is `https://MY_HOST/ollama`. I can access the Ollama API and get the response using curl:
```sh
curl https://MY_HOST/ollama/api/chat -d '{
  "model": "llama3",
  "stream": "false",
  "messages": [
    {
      "role": "user",
      "content": "who are you?"
    }
  ]
}'
```

However, I couldn't set the `https://MY_HOST/ollama` to the `Client`. The final url alway be parsed to `https://MY_HOST:443` without the path `ollama`.

```python
# This is the default value
client = Client(host="http://localhost:11434")
client._client.base_url  # URL('http://localhost:11434')

# This is the actual behavior
client = Client(host="https://MY_HOST/ollama")
client._client.base_url  # URL('https://MY_HOST')

# This is what I expected
client = Client(host="https://MY_HOST/ollama")
client._client.base_url  # URL('https://MY_HOST/ollama')
```

------

If this is reasonable, I could provide a PR for this feature.

bplunkert:
This fix for #113 allows instantiating a client with a custom URL path as part of the host argument.

Resolves #244 
Resolves #113


HorseCheng:
Someone please approve the PR and merge the code. 🥺

jason-heo:
Resolves #244 as well.

I'd love to see this PR get merged.

ollama/_client.py
```diff
@@ -725,5 +725,6 @@ def _parse_host(host: Optional[str]) -> str:
   split = urllib.parse.urlsplit('://'.join([scheme, hostport]))
   host = split.hostname or '127.0.0.1'
   port = split.port or port
+  path = split.path or ''
 
-  return f'{scheme}://{host}:{port}'
+  return f'{scheme}://{host}:{port}{path}'
```
mxyng, line 730: I would prefer this instead. Otherwise tests need to be updated.

```python
  if path := split.path.strip('/'):
    return f'{scheme}://{host}:{port}/{path}'

  return f'{scheme}://{host}:{port}'
```

mxyng (commented):
Thanks for the PR. I can merge this is you add some doc tests. Something like this should be sufficient:

```python
  >>> _parse_host('example.com/path')
  'http://example.com:11434/path'
  >>> _parse_host('example.com:56789/path')
  'http://example.com:56789/path'
  >>> _parse_host('https://example.com:56789/path')
  'https://example.com:56789/path'
  >>> _parse_host('example.com:56789/path/')
  'http://example.com:56789/path'
```

bplunkert:
thanks @mxyng I've added your suggestions to the pull request.

ollama/_client.py
```diff
@@ -725,5 +733,9 @@ def _parse_host(host: Optional[str]) -> str:
   split = urllib.parse.urlsplit('://'.join([scheme, hostport]))
   host = split.hostname or '127.0.0.1'
   port = split.port or port
+  path = split.path or ''
```
mxyng, line 736: This is unnecessary now since `path` is assigned in the if-block below

bplunkert: oops. fixed!

mxyng (approved):
Awesome thanks!

ollama/_client.py
```diff
@@ -711,6 +711,14 @@ def _parse_host(host: Optional[str]) -> str:
...
   >>> _parse_host('example.com:56789/')
   'http://example.com:56789'
+  >>> _parse_host('example.com/path')
+  'http://example.com:11434/path'
+  >>> _parse_host('example.com:56789/path')
+  'http://example.com:56789/path'
+  >>> _parse_host('https://example.com:56789/path')
+  'https://example.com:56789/path'
+  >>> _parse_host('example.com:56789/path/')
+  'http://example.com:56789/path/'
```
mxyng, line 721: Oops the expected output should not end in a `/`

PR #179, opened 2024-06-05
Simple Example

royjhan:
Could show cpu vs gpu proportion calculations if helpful but just pushing the response for now

### DIFF
--- a/examples/ps/main.py
+++ b/examples/ps/main.py
@@ -10,10 +10,10 @@ size = response['models'][0]['size']
 size_vram = response['models'][0]['size_vram']
 
 if size == size_vram:
-    print(f'{name}: 100% GPU')
+  print(f'{name}: 100% GPU')
 elif not size_vram:
-    print(f'{name}: 100% CPU')
+  print(f'{name}: 100% CPU')
 else:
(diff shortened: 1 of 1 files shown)


examples/ps/main.py
```diff
@@ -0,0 +1,4 @@
+from ollama import ps
+
+response = ps()
+print(response)
```
mxyng, line 4: This example should be a bit more involved. To really show ps, you should load a model by calling chat or generate

examples/ps/main.py
```diff
@@ -0,0 +1,19 @@
+from ollama import ps, pull, chat
+
+pull('mistral')
+chat('mistral', messages=[{'role': 'user', 'content': 'Pick a number'}])
```
mxyng, line 4: it might be useful to output the output content

examples/ps/main.py
```diff
@@ -0,0 +1,20 @@
+from ollama import ps, pull, chat
+
+pull('mistral')
```
mxyng, line 3: nit: pull blocks without feedback since it doesn't stream which leaves the user wondering what's going on

mxyng (approved):

PR #181, opened 2024-06-05
Add type overloads to methods

royjhan:
In reference to #132

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -111,7 +111,8 @@ class Client(BaseClient):
     images: Optional[Sequence[AnyStr]] = None,
     options: Optional[Options] = None,
     keep_alive: Optional[Union[float, str]] = None,
-  ) -> Mapping[str, Any]: ...
+  ) -> Mapping[str, Any]:
+    ...
 
   @overload
   def generate(
@@ -127,7 +128,8 @@ class Client(BaseClient):
     images: Optional[Sequence[AnyStr]] = None,
     options: Optional[Options] = None,
     keep_alive: Optional[Union[float, str]] = None,
-  ) -> Iterator[Mapping[str, Any]]: ...
+  ) -> Iterator[Mapping[str, Any]]:
+    ...
 
   def generate(
     self,
@@ -184,7 +186,8 @@ class Client(BaseClient):
     format: Literal['', 'json'] = '',
     options: Optional[Options] = None,
     keep_alive: Optional[Union[float, str]] = None,
-  ) -> Mapping[str, Any]: ...
+  ) -> Mapping[str, Any]:
+    ...
 
   @overload
   def chat(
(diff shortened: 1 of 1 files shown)


ollama/_client.py
```diff
@@ -97,6 +97,38 @@ def _request_stream(
...
 
+  @overload
+  def generate(
+    self,
+    model: str = '',
+    prompt: str = '',
+    system: str = '',
+    template: str = '',
+    context: Optional[Sequence[int]] = None,
+    stream: bool = False,
```
mxyng, line 108: This is incorrect since it creates an overlap between the two overloads

```suggestion
    stream: Literal[False] = False,
```

ollama/_client.py
```diff
@@ -97,6 +97,38 @@ def _request_stream(
...
+
+  @overload
+  def generate(
+    self,
+    model: str = '',
+    prompt: str = '',
+    system: str = '',
+    template: str = '',
+    context: Optional[Sequence[int]] = None,
+    stream: bool = True,
```
mxyng, line 124:
```suggestion
    stream: Literal[True] = True,
```

mxyng (changes requested):

ollama/_client.py
```diff
@@ -431,11 +569,33 @@ async def generate(
...
+    format: Literal['', 'json'] = '',
+    options: Optional[Options] = None,
+    keep_alive: Optional[Union[float, str]] = None,
+  ) -> AsyncIterator[Mapping[str, Any]]: ...
+
+  async def chat(
+    self,
+    model: str = '',
+    messages: Optional[Sequence[Message]] = None,
+    stream: Literal[False] = False,
```
mxyng, line 598:
```suggestion
    stream: bool = False,
```

ollama/_client.py
```diff
@@ -97,6 +97,38 @@ def _request_stream(
...
+    system: str = '',
+    template: str = '',
+    context: Optional[Sequence[int]] = None,
+    stream: Literal[False] = False,
+    raw: bool = False,
+    format: Literal['', 'json'] = '',
+    images: Optional[Sequence[AnyStr]] = None,
+    options: Optional[Options] = None,
+    keep_alive: Optional[Union[float, str]] = None,
+  ) -> Mapping[str, Any]: ...
```
mxyng, line 114: this is minor but it'd be better if `...` was on a newline. it's easier to read and harder to miss

mxyng (approved):

royjhan:
poetry doesn't let me move the ellipsis to a new line, lmk and I can just revert and merge

mxyng:
> poetry doesn't let me move the ellipsis to a new line, lmk and I can just revert and merge

ok let's go with the previous, inline style :(

Issue #184, opened 2024-06-07
Interactive Mode


p-vdp:
It would be useful to have an interactive mode accessed via CLI and the REST API (see also #183) , something like this:

```
$ python -m ollama --endpoint http://remotehost:11434 --model llama3 --interactive

[Connected to endpoint: http://remotehost:11434]
[Model: llama3]

$ Why is the sky blue?

The sky appears blue because of a phenomenon called scattering, which occurs when sunlight interacts 
with tiny molecules of gases in the Earth's atmosphere.
```

mxyng:
As mentioned in #185, a CLI is be out of scope of this repo

Issue #183, opened 2024-06-07
Command Line Interface


p-vdp:
I'm running ollama on a remote server and would like to interact with it from another machine via CLI and the REST API.

This should be pretty straightforward to do with a custom Client object in `__main__.py`. Usage from a terminal would be something like this:

`python -m ollama --endpoint http://remotehost:11434 --model llama3 --message "Why is the sky blue?"`

I'm interested in implementing this. Do folks have a preference for CLI parameter handling? I prefer [absl](https://abseil.io/docs/python/quickstart).

synacktraa:
I prefer [Cyclopts](https://www.reddit.com/r/Python/comments/18hn2t1/cyclopts_a_cli_library_that_fixes_13_annoying/)

mxyng:
As mentioned in #185, a CLI is be out of scope of this repo

PR #185, opened 2024-06-07
CLI proof of concept

p-vdp:
See #183 #184

### DIFF
--- a/ollama/__main__.py
+++ b/ollama/__main__.py
@@ -1,0 +1,32 @@
+from absl import app, flags
+
+from ollama._client import Client
+
+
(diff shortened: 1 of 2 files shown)


Omega-Centauri-21 (commented):
Don't we already have an Ollama CLI? Why reinvent the wheel with python? 
Also Go is faster than Python ... so whats the real use of this PoC?

mxyng:
Thanks for the contribution. As another contributor has pointed out, a Python-based doesn't really make sense since Ollama CLI already exists. This library is meant to be a thin wrapper around the Ollama API. As such, a CLI would be out of scope.

PR #201, opened 2024-07-01
fix: PydanticUserError: Please use `typing_extensions.TypedDict` inst…

tang2087:
…ead of `typing.TypedDict` on Python < 3.12.


ParthSareen:
Should already be addressed through the recent pydantic integrations - will be fixed in the new release :)

PR #207, opened 2024-07-06
Adding an example how you can ask several models something.

gonaumov:
I am offering an example how you can ask a question several modells, display their responses and store them in a text file.


ParthSareen:
Hi @gonaumov! While this is a cool display of Ollama's capabilities we're trying to keep our examples basic for now. Thank you for this though!

PR #208, opened 2024-07-11
Support `api/embed`

royjhan:


mxyng (approved):

ollama/_client.py
```diff
@@ -245,6 +245,29 @@ def chat(
...
+      json={
+        'model': model,
+        'input': input,
+        'truncate': truncate,
+        'options': options or {},
+        'keep_alive': keep_alive,
+      },
+    ).json()
+
   def embeddings(
```
mxyng, line 271: this can be marked deprecated with https://docs.python.org/3/library/exceptions.html#DeprecationWarning

royjhan: will make note of this in docs pr

PR #213, opened 2024-07-16
integrate tool calls

joshyan1:


BruceMacD (commented):
This looks good to me, pending a check from @mxyng. I believe the tests wont pass until we release the new ollama version.

mxyng (approved):
lgtm. it'd be nice to define anonymous classes for the functions but there's no way of doing that while also making the linter happy.

PR #215, opened 2024-07-17
add insert support to generate endpoint

royjhan:
Update fill-in-the-middle support to use the suffix field.

Implements: https://github.com/ollama/ollama/pull/5207


examples/fill-in-middle/main.py
```diff
@@ -1,22 +1,9 @@
...
-    """ '''
-
-suffix = """
-    return result
-"""
-
-
 response = generate(
-  model='codellama:7b-code',
-  prompt=f'<PRE> {prefix} <SUF>{suffix} <MID>',
```
mxyng, line 13: This example should remain the same except use prompt and suffix instead of templating prompt with f-string

mxyng (changes requested):

mxyng (approved):

PR #227, opened 2024-07-23
Update `Message` annotations to support tool calls

Shulyaka:
This is to stay in line with the tool call example.


mxyng (approved):
Thanks for the change!

PR #234, opened 2024-07-26
Add Tool Registry Feature

synacktraa:
## Description
This PR introduces a new `ToolRegistry` class to the Ollama library. The Tool Registry is designed to manage various tools, including functions (both synchronous and asynchronous), `pydantic.BaseModel`, `typing.TypedDict`, and `typing.NamedTuple`. It provides functionality to generate schemas for these tools, which are essential for LLM tool-calling, and allows for the invocation of tools using their metadata.

## Features
- Create and manage a registry of tools
- Support for multiple tool types: functions, pydantic models, TypedDict, and NamedTuple
- Generate schemas for registered tools
- Invoke tools using their metadata (name & raw arguments) -- Handles almost all edge cases
- Option to override existing tools with the same name

## Changes I've made:
- Added a new directory `tool_calling` which stores all functionality of tool registry.
- Added unit tests to cover all major functionality of the `ToolRegistry` class. (All tests passed)
- Updated _types.py
- Updated tools example in examples directory

### Creating a tool registry

```python
import ollama

registry = ollama.ToolRegistry()
```

### Adding tools to the registry instance

```python
import json
import asyncio
from typing import Literal, TypedDict

registry = ollama.ToolRegistry()

@registry.register
def get_flight_times(departure: str, arrival: str) -> str:
  """
  Get flight times.
  :param departure: Departure location code
  :param arrival: Arrival location code
  """
  flights = {
    'NYC-LAX': {'departure': '08:00 AM', 'arrival': '11:30 AM', 'duration': '5h 30m'},
    'LAX-NYC': {'departure': '02:00 PM', 'arrival': '10:30 PM', 'duration': '5h 30m'},
    'LHR-JFK': {'departure': '10:00 AM', 'arrival': '01:00 PM', 'duration': '8h 00m'},
    'JFK-LHR': {'departure': '09:00 PM', 'arrival': '09:00 AM', 'duration': '7h 00m'},
    'CDG-DXB': {'departure': '11:00 AM', 'arrival': '08:00 PM', 'duration': '6h 00m'},
    'DXB-CDG': {'departure': '03:00 AM', 'arrival': '07:30 AM', 'duration': '7h 30m'},
  }

  key = f'{departure}-{arrival}'.upper()
  return json.dumps(flights.get(key, {'error': 'Flight not found'}))

@registry.register
class User(TypedDict): # It can also be `pydantic.BaseModel`, or `typing.NamedTuple`.
    """
    User Information
    :param name: Name of the user
[...]

PR #238, opened 2024-07-29
Autogen Tool, ToolFunction, Parameters, Property schema when opt-in to using type-hinted function params

anthonywu:
# 😩 Problem

In the [API docs](https://github.com/ollama/ollama/blob/main/docs/api.md#chat-request-with-tools) for passing in `tools=` spec sequence to the chat model, the docs suggest developers hand-craft the `Tool`, `ToolFunction`, `Parameters`, `Property` schema'ed docs to the `tools=` arg. This is very repetitive.

For example, building up this JSON is laborious, given we already have defined the tool calling function.

```json
{
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather for a location",
        "parameters": { 😩
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The location to get the weather for, e.g. San Francisco, CA"
            }, 😩
            "format": {
              "type": "string",
              "description": "The format to return the weather in, e.g. 'celsius' or 'fahrenheit'",
              "enum": ["celsius", "fahrenheit"] 😩
            }
          },
          "required": ["location", "format"] 😩
        }
      }
```

# Improvement Proposal

I propose we can solve it by giving users an option to type hint their functions and get back an automatic schema.

The scope is:

- **must only** use Python standard library and native `type` `typing` helpers
- your literal function name in `.py` is the function name in the spec, no need to type it twice
- `function` docstring **is** the `Function` `description`, this should be a wide-spread developer habit, some possibility to inject prompts here
- parameter description provided by the second arg of [`Annotated` special typing form](https://docs.python.org/3/library/typing.html#typing.Annotated)
- required/optional state is hinted by [`Optional` annotation](https://docs.python.org/3/library/typing.html#typing.Optional) or classic keyword args with default values
- the documented types are supported
    - `bool` -> `"boolean"`
    - `float` -> `"number"`
    - `int` -> `"integer"`
    - `str` -> `"string"`
    - `Sequence` -> `"array"`

The user expectation is:

- this is **optional**, if you want to hand craft the spec, go for it
- the generated spec is just plain JSON, you can edit it however you wish if the auto generated spec is not exactly what you need

# Example

```py
@ollama.annotated_tool
def some_tool_function(
    expr_1: t.Annotated[str, "first simple str value"],
    expr_2: t.Annotated[str, "second simple str value"],
    req_int_arg: t.Annotated[int, "a required integer"],
    req_float_arg: t.Annotated[float, "any real number"],
    req_list_arg_1: t.Annotated[list, "any builtin list"],
    req_list_arg_2: t.Annotated[t.List[t.Any], "any typed List"],
    req_enum_1: t.Annotated[ExampleEnum, "required foo bar or baz"],
    opt_arg_1: t.Annotated[t.Optional[str], "an optional string"] = None,
    opt_enum_1: t.Annotated[t.Optional[ExampleEnum], "optional foo bar or baz"] = None,
    opt_enum_2: ExampleEnum = ExampleEnum.FOO,
    opt_builtin_str_1: str = "foobar",
    opt_builtin_int_1: int = 1e6,
    opt_builtin_float_1: float = 1.0,
    opt_builtin_list_1: list = None,
):
    """a test case for Annotating a tool function's parameters.
    the docstring of the function is the Function.description."""
    return None
```

`some_tool_function.tool_schema` would provide this, usable as an item in `tools=` list

```json
{
  "type": "function",
  "function": {
    "name": "example_1",
    "description": "a test case for Annotating a tool function's parameters.\n    the docstring of the function is the Function.description.",
    "parameters": {
      "type": "object",
      "required": [
        "expr_2",
        "expr_1",
        "req_int_arg",
        "req_list_arg_2",
        "req_enum_1",
        "req_float_arg",
        "req_list_arg_1"
      ],
      "properties": {
        "expr_1": {
          "type": "string",
          "description": "first simple str value",
          "is_optional": false,
          "enum": null
        },
        "expr_2": {
          "type": "string",
          "description": "second simple str value",
          "is_optional": false,
          "enum": null
        },
        "req_int_arg": {
          "type": "integer",
          "description": "a required integer",
          "is_optional": false,
          "enum": null
        },
        "req_float_arg": {
          "type": "number",
          "description": "any real number",
          "is_optional": false,
          "enum": null
        },
        "req_list_arg_1": {
          "type": "array",
          "description": "any builtin list",
          "is_optional": false,
          "enum": null
        },
        "req_list_arg_2": {
          "type": "array",
          "description": "any typed List",
          "is_optional": false,
          "enum": null
        },
        "req_enum_1": {
          "type": "string",
          "description": "required foo bar or baz",
          "is_optional": false,
          "enum": [
            "foo",
            "bar",
            "baz"
          ]
        },
        "opt_arg_1": {
          "type": "string",
          "description": "an optional string",
          "is_optional": true,
          "enum": null
        },
        "opt_enum_1": {
          "type": "string",
          "description": "optional foo bar or baz",
          "is_optional": true,
          "enum": [
            "foo",
            "bar",
            "baz"
          ]
        },
        "opt_enum_2": {
          "type": "string",
          "description": "opt enum 2",
          "is_optional": true,
          "enum": [
            "foo",
            "bar",
            "baz"
          ]
        },
        "opt_builtin_str_1": {
          "type": "string",
          "description": "opt builtin str 1",
          "is_optional": true,
          "enum": null
        },
        "opt_builtin_int_1": {
          "type": "integer",
          "description": "opt builtin int 1",
          "is_optional": true,
          "enum": null
        },
        "opt_builtin_float_1": {
          "type": "number",
          "description": "opt builtin float 1",
          "is_optional": true,
          "enum": null
        },
        "opt_builtin_list_1": {
          "type": "array",
          "description": "opt builtin list 1",
          "is_optional": true,
          "enum": null
        }
      }
    }
  }
}
```

# Out of Scope

Not attempting to scope creep to:

- consistency with OpenAI Python SDK's similar implementation
- not using `pydantic` so to not obligate `pip install ollama` users to also install `pydantic` (see comment in https://github.com/ollama/ollama-python/pull/234#issuecomment-2253032468)
- make this a stand alone `pip install`able library, I think the implementation is deeply coupled to `ollama`'s `_types.py` implementation. I am open to doing this implementation as a separate library if the maintainers believe this syntactic sugar is out of scope.

# Related

- #234 "Add Tool Registry Feature" is highly related, however it was closed for recommendation to package as a third party library. My implementation here adds no additional dependencies and remains entirely optional.
- There was a related attempt to solve this problem: https://github.com/robocorp/llmfoo but that attempt sends the source of the function to OpenAI API to let the LLM generate the tool schema. My solution here is entirely local and relies on the Python 3 standard type utils only, 100% type reflection.

### DIFF
--- a/examples/tool-functions/main.py
+++ b/examples/tool-functions/main.py
@@ -1,0 +1,74 @@
+import typing as t
+import ollama
+
+from ollama._toolfuncs import annotated_tool  # Pull Request coming soon
+
+
+def lookup_function(function_name: str):
+    return globals()[function_name]
+
+
+# Example Function 1
+@annotated_tool
+def get_current_weather(city: t.Annotated[str, "The name of the city"]):
+    # mock response
+    return 75.0, "Farenheit"
+
+
+def run_example_1():
+    response_1 = ollama.chat(
+        model="llama3.1",
+        messages=[{"role": "user", "content": "What is the weather in San Francisco?"}],
+        tools=[get_current_weather.tool_schema],
+    )
+
+    fn_call_1 = response_1["message"]["tool_calls"][0]["function"]
+    fn_resp_1 = lookup_function(fn_call_1["name"])(city=fn_call_1["arguments"]["city"])
+    assert fn_resp_1 == (75.0, "Farenheit")
+    print("SUCCESS: get_current_weather is called")
+
+
+run_example_1()
+
+
+# Example Function 2
+@annotated_tool
+def sort_fruit_names(
+    fruit_names: t.Annotated[t.List[str], "a list of fruit names"]
+) -> t.List:
+    return sorted(fruit_names)
+
+
+def run_example_2():
+    response_2 = ollama.chat(
+        model="llama3.1",
+        messages=[
+            {
+                "role": "user",
+                "content": "Sort the fruit names strawberry, kiwi, banana, pineapple, passion fruit, mango",
+            }
+        ],
+        tools=[
+            get_current_weather.tool_schema,  # not expecting a match
+            sort_fruit_names.tool_schema,  # expecting a match
+        ],
+    )
+    print(response_2["message"]["tool_calls"])
+    assert len(response_2["message"]["tool_calls"]) == 1
+    fn_call_2 = response_2["message"]["tool_calls"][0]["function"]
+    fruit_names_arg = fn_call_2["arguments"]["fruit_names"]
+    print(f"model found {len(fruit_names_arg)} fruit names: {fruit_names_arg}")
+    fn_resp_2 = lookup_function(fn_call_2["name"])(fruit_names=fruit_names_arg)
+    print(fn_resp_2)
+    assert fn_resp_2 == [
+        "banana",
+        "kiwi",
+        "mango",
+        "passion fruit",
+        "pineapple",
+        "strawberry",
+    ]
+    print("SUCCESS: sort_fruit_names is called")
+
+
+run_example_2()
--- a/ollama/__init__.py
+++ b/ollama/__init__.py
@@ -8,6 +8,7 @@ from ollama._types import (
   RequestError,
   ResponseError,
 )
+from ollama._toolfuncs import annotated_tool
 
 __all__ = [
   'Client',
--- a/ollama/_toolfuncs.py
+++ b/ollama/_toolfuncs.py
@@ -1,0 +1,157 @@
+import inspect
+from enum import EnumMeta
+from types import NoneType
+from typing import (
+    Annotated,
+    Dict,
+    List,
+    Sequence,
+    TypedDict,
+    Union,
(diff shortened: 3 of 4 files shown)


ParthSareen:
Kind of related but I fixed this because I also wasn't a fan of building tool descriptions myself. The prompt works with even 8b models locally. https://github.com/Extensible-AI/DAGent/blob/8dd76ffe7c1bdd57ebfcf67a859f25d196632233/dagent/src/dagent/base_functions.py#L16-L45

anthonywu:
Withdrawing PR. Made this available via its own repo: https://github.com/anthonywu/ollama-tool-register and `pip install ollama-tool-register`. My further developments will be in this separate artifact. Collaborators welcome over at my repo.

PR #262, opened 2024-08-23
IPv6 support

jbinder:
Support connecting to hosts via an IPv6 address.

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -1,3 +1,4 @@
+import ipaddress
 import os
 import io
 import json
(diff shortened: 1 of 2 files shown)


ollama/_client.py
```diff
@@ -1013,4 +1015,8 @@ def _parse_host(host: Optional[str]) -> str:
   if path := split.path.strip('/'):
     return f'{scheme}://{host}:{port}/{path}'
 
+  # Fix missing square brackets for IPv6 from urlsplit
+  if hostport.startswith('['):
```
mxyng, line 1019: I would prefer something more explicit with the `ipaddress` package:

```python
host = split.hostname or '127.0.0.1'
try:
    if isinstance(ipaddress.ip_address(host), ipaddress.IPv6Address)):
        host = f'[{host}]'
except ValueError:
    ...
```

tests/test_client.py
```diff
@@ -992,3 +992,31 @@ async def test_async_client_create_blob_exists(httpserver: HTTPServer):
   with tempfile.NamedTemporaryFile() as blob:
     response = await client._create_blob(blob.name)
     assert response == 'sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+
+
+def test_parse_host(httpserver: HTTPServer):
```
mxyng, line 997: this is unnecessary since these are tested in the doc test

ollama/_client.py
```diff
@@ -995,6 +995,8 @@ def _parse_host(host: Optional[str]) -> str:
   'https://example.com:56789/path'
   >>> _parse_host('example.com:56789/path/')
   'http://example.com:56789/path'
+  >>> _parse_host('[0001:002:003:0004::1]')
```
mxyng, line 998: maybe a few more with scheme, port, and path but particularly port

tests/test_client.py
jbinder: Ah thanks, I missed that!

jbinder:
@mxyng Thanks a lot for your feedback!

PR #266, opened 2024-08-28
_stream in async client raises RuntimeError when processing HTTP errors

Oneirag:
When response received in AsyncClient._stream has an HTTP error, the process raises a RuntimeError("Attempted to call a sync iterator on an async stream.") due to the use of e.response.read() instead of await e.response.aread()

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -495,7 +495,7 @@ class AsyncClient(BaseClient):
         try:
           r.raise_for_status()
         except httpx.HTTPStatusError as e:
-          e.response.read()
+          await e.response.aread()
           raise ResponseError(e.response.text, e.response.status_code) from None
 
         async for line in r.aiter_lines():


mxyng (approved):
Thanks for the fix!

WikiLucas00:
Hello @mxyng I don't think this fix is included in the latest release. Do you have a estimated date for the next release please?

mxyng:
I just created a new release. Thanks for the reminder.

WikiLucas00:
You're welcome! Thanks a lot for your responsiveness :pray:

chrisaww:
context: no claims to be an expert in this area.
getting  error when trying to run async with pydantic framework
 raise exceptions.UnexpectedModelBehavior(
        f'Exceeded maximum retries ({self._max_result_retries}) for result validation'
    )
pydantic_ai.exceptions.UnexpectedModelBehavior: Exceeded maximum retries (1) for result validation
also notice Langchain has a similar issue raised

PR #276, opened 2024-09-10
add pydantic for validation and serialization

mxyng:
while the interface should be backwards compatible, i.e. most tests and examples do not need any changes, this change wants a minor version bump

examples should be updated in a following change to use pydantic models where appropriate


ollama/_client.py
```diff
@@ -450,126 +524,184 @@ def _create_blob(self, path: Union[str, Path]) -> str:
...
+    )
 
-  def ps(self) -> Mapping[str, Any]:
-    return self._request('GET', '/api/ps').json()
+  def ps(self) -> ProcessResponse:
+    return self._request(
+      ProcessResponse,
+      'GET',
+      '/api/ps',
+    )
```
ParthSareen, line 584: Need the model dump here as well
```suggestion
    ).model_dump(exclude_none=True)
```

ollama/_client.py
```diff
@@ -888,60 +1021,70 @@ async def upload_bytes():
...
-  'YWJj'
-  >>> _encode_image(b'YWJj')
-  'YWJj'
-  """
+  async def ps(self) -> ProcessResponse:
+    return await self._request(
+      ProcessResponse,
+      'GET',
+      '/api/ps',
+    )
```
ParthSareen, line 1075:
```suggestion
        response = await self._request(
      ProcessResponse,
      'GET',
      '/api/ps',
    )
    return response.model_dump(exclude_none=True)
```

ParthSareen: Gonna add an async example for ps as well

ollama/_client.py
ParthSareen: Going to tackle these comments in new PR

ParthSareen:
See https://github.com/ollama/ollama-python/pull/311 for pydantic type class fixes and serialization

ParthSareen (approved):
Woooo Pydantic!!! 🥳 🎉

ollama/_client.py
```diff
@@ -48,16 +86,15 @@ def __init__(
...
     self._client = client(
       base_url=_parse_host(host or os.getenv('OLLAMA_HOST')),
       follow_redirects=follow_redirects,
       timeout=timeout,
-      headers=headers,
+      headers={
+        'Content-Type': 'application/json',
+        'Accept': 'application/json',
+        'User-Agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}',
+      }.update(headers or {}),
```
ParthSareen, line 97: `.update()` is going to return a `None` which means no headers will get passed in

ParthSareen (changes requested):

ParthSareen (approved):
🫡

PR #282, opened 2024-09-17
add OpenAI example in README.md.

fujitatomoya:


fujitatomoya (commented):
this would be good information to introduce OpenAI compatible API explicitly to Ollama users in the 1st place.

fujitatomoya:
@jmorganca could you take a look at this?

btw, i am happy to review PRs if you guys do not have bandwidth. please let me know!

kth8:
The openai-python library is completely separate from the ollama-python that uses a different API so it doesn't seem appropriate to include that here. If people want to use the OpenAI API with Ollama, there are already instructions to do so in the main repo. https://github.com/ollama/ollama/blob/main/docs/openai.md

fujitatomoya:
@kth8 thanks for the feedback. Ah i see, i did not know that https://github.com/ollama/ollama/blob/main/docs/openai.md exist, but as you mentioned that would be the appropriate place to maintain OpenAI compatibility information. all that said, i will go ahead to close this one.

PR #299, opened 2024-10-13
Enable subpath-mounted ollama and basic auth

jamesandariese:
* Modifies _parse_host to preserve more info and return a URL which is compatible with relative paths.

  * Enables subpaths, used, in the next bullet point.
  * Enables basic auth, query params, and fragments

* Modifies all API calls to be relative so that a subpath-mounted ollama may be called.
  * Enables instance-specific paths (e.g., https://ollama.example.com/instance-00000000-1111-2222-3333-444444444444)

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -162,7 +162,7 @@ class Client(BaseClient):
 
     return self._request_stream(
       'POST',
-      '/api/generate',
+      'api/generate',
       json={
         'model': model,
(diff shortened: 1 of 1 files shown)


ParthSareen:
IMO it doesn't make a lot of sense to package the two together - especially for auth. Auth should be done through headers and proxies if anything. Happy to reconsider if needed!

PR #311, opened 2024-11-06
Pydantic Fixes and Tests

ParthSareen:
Improper class name was set - moving from `BaseModel` to `SubscriptableBaseModel`

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -97,6 +97,7 @@ class BaseGenerateRequest(BaseStreamableRequest):
 class Image(BaseModel):
   value: Union[FilePath, Base64Str, bytes]
 
+  # This overloads the `model_dump` method and returns values depending on the type of the `value` field
   @model_serializer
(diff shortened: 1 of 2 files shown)


ollama/_types.py
```diff
@@ -333,7 +334,7 @@ class ModelDetails(SubscriptableBaseModel):
 
 
 class ListResponse(SubscriptableBaseModel):
-  class Model(BaseModel):
+  class Model(SubscriptableBaseModel):
```
ParthSareen, line 337: Didn't end up gutting this out - did a quick mock to see what it would look like and I think it's not worth the extra complexity just yet. I think there is some server and pattern work to be done before we can have a base class for model. 

I'd ideally want to be able to share this class a lot more freely with those consistencies. With that, I think it's okay for it to be out of scope for this and your PR.

mxyng (approved):

PR #313, opened 2024-11-07
Fixing empty header + ensuring security

ParthSareen:
To fix https://github.com/ollama/ollama-python/pull/276#discussion_r1831885785

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -90,11 +90,13 @@ class BaseClient:
       base_url=_parse_host(host or os.getenv('OLLAMA_HOST')),
       follow_redirects=follow_redirects,
       timeout=timeout,
+      # Lowercase all headers to ensure override
       headers={
-        'Content-Type': 'application/json',
-        'Accept': 'application/json',
-        'User-Agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}',
-      }.update(headers or {}),
+        **{k.lower(): v for k, v in (headers or {}).items()},
+        'content-type': 'application/json',
+        'accept': 'application/json',
+        'user-agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}',
+      },
       **kwargs,
     )
 
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -91,12 +91,7 @@ class BaseClient:
       follow_redirects=follow_redirects,
       timeout=timeout,
       # Lowercase all headers to ensure override
-      headers={
-        **{k.lower(): v for k, v in (headers or {}).items()},
-        'content-type': 'application/json',
-        'accept': 'application/json',
-        'user-agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}',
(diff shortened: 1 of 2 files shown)


ollama/_client.py
```diff
@@ -90,11 +90,13 @@ def __init__(
       base_url=_parse_host(host or os.getenv('OLLAMA_HOST')),
       follow_redirects=follow_redirects,
       timeout=timeout,
+      # Lowercase all headers to ensure override
       headers={
-        'Content-Type': 'application/json',
-        'Accept': 'application/json',
-        'User-Agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}',
-      }.update(headers or {}),
+        **{k.lower(): v for k, v in (headers or {}).items()},
```
ParthSareen, line 95: Had to run `.lower()` since python would keep both keys but during a http request they would double - so jank

mxyng: instead of lowering `headers` keys, you can simplify this slightly by lowering the final dict before passing to httpx. that will be less error prone since the client-set headers will also be case insensitive 

```python
self._client = client(
    headers={
        k.lower(): v for k, v in {
            **(headers or {}),
            'Content-Type': 'application/json',
            'Accept': 'application/json',
            'User-Agent': f'...',
        }
    },
)
```

tests/test_client.py
```diff
@@ -968,3 +968,19 @@ async def test_async_client_copy(httpserver: HTTPServer):
   client = AsyncClient(httpserver.url_for('/api/copy'))
   response = await client.copy('dum', 'dummer')
   assert response['status'] == 'success'
+
+
+def test_headers():
+  client = Client()
+  assert client._client.headers['content-type'] == 'application/json'
+  assert client._client.headers['accept'] == 'application/json'
+  assert 'ollama-python' in client._client.headers['user-agent']
```
mxyng, line 977:
```suggestion
  assert client._client.headers['user-agent'].startswith('ollama-python/')
```

ollama/_client.py
ParthSareen: The merging of the dicts inside the `.lower` would be case based which means that we'd still get unwanted behavior. Not sure how what the new dict would look like since there would be key collision after the `.lower()`

ollama/_client.py
```diff
@@ -90,11 +90,8 @@ def __init__(
       base_url=_parse_host(host or os.getenv('OLLAMA_HOST')),
       follow_redirects=follow_redirects,
       timeout=timeout,
-      headers={
-        'Content-Type': 'application/json',
-        'Accept': 'application/json',
-        'User-Agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}',
-      }.update(headers or {}),
+      # Lowercase all headers to ensure override
+      headers={k.lower(): v for k, v in {**(headers or {}), 'Content-Type': 'application/json', 'Accept': 'application/json', 'User-Agent': f'ollama-python/{__version__} ({platform.machine()} {platform.system().lower()}) Python/{platform.python_version()}'}.items()},
```
mxyng, line 94: this should be formatted for readability with each key/value should have its own line

mxyng (approved):

PR #315, opened 2024-11-07
Restrict httpx version to not introduce breaking changes

ParthSareen:
- Restrict the versions for httpx between 0.27 and <0.28
- Possible breaking changes can be introduced even with a minor bump for such a young project.

Inspired from https://github.com/ollama/ollama-python/pull/97#issuecomment-2023394428
Closes: https://github.com/ollama/ollama-python/pull/265


ParthSareen:
Closes https://github.com/ollama/ollama-python/pull/265

pyproject.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/ollama/ollama-python"
 
 [tool.poetry.dependencies]
 python = "^3.8"
-httpx = "^0.27.0"
+httpx = ">=0.27.0,<0.28.0"
```
mxyng, line 13: this has the same semantics so it's unnecessary https://python-poetry.org/docs/dependency-specification/#caret-requirements

mxyng (approved):

PR #321, opened 2024-11-11
Passing Functions as Tools

ParthSareen:
## Features:
- Pass in python functions as tools into `.chat(tools=[python_func,...])` 🥳 
- Follows Google Python Docstring Styleguide - See [Google Docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)
- Lenient on docstring specification - should handle unusual cases elegantly
- Can still use a mixture of Python functions as tools and JSON values - feature is purely additive.

Examples to come on follow up for both Pydantic and Functions as Tools


ollama/_types.py
```diff
@@ -229,12 +230,14 @@ class Function(SubscriptableBaseModel):
     description: str
 
     class Parameters(SubscriptableBaseModel):
-      type: str
+      type: Union[str, List[str]]
```
ParthSareen, line 233: Appropriate change was made serverside as well - will discuss implications there https://github.com/ollama/ollama/pull/7613

ollama/_utils.py
```diff
@@ -0,0 +1,188 @@
...
+# Type compatibility layer
+if sys.version_info >= (3, 10):
+  from types import UnionType
+
+  def is_union(tp: Any) -> bool:
+    return get_origin(tp) in (Union, UnionType)
+else:
+
+  def is_union(tp: Any) -> bool:
+    return get_origin(tp) is Union
```
ParthSareen, line 18: When types similar to `int | str` or `<type> | <type> | ...` are used in newer python that maps to `UnionType` which doesn't exist in Python < 3.10. 

The parser has logic to check for unions but older python versions would fail as that typing structure does not exist (Python 3.14 is merging these types as well - https://github.com/python/cpython/pull/105511)

After lots of testing the best I could come up with to keep everything happy was setup the function given python version. It's not my favorite addition but the user experience will be a drastic improvement.

tests/test_utils_legacy.py
```diff
@@ -0,0 +1,77 @@
...
+    y: str,
+    z: Sequence[int],
+    w: Mapping[str, int],
+    d: Dict[str, int],
+    s: Set[int],
+    t: Tuple[int, str],
+    l: List[int],  # noqa: E741
+    o: Union[int, None],  # Changed Optional to Union
+  ) -> Union[Mapping[str, int], str, None]:
+    """
```
ParthSareen, line 52: Something important to note here is that if  `from __future__ import annotations` is used the parser breaks as it is currently parsing types directly and not a string. 

Specifically, it breaks for nested types like Union[..., ...] and Mapping[..., ...]. 

I could write a string parser if we think this is a big deal but it will be a bit of a time sink for sure. Good to decide what the support level should be python 3.9 and how opinionated we need to be. I think in this case particular the time sink is not worth the support but open to hearing thoughts.

tests/test_utils.py
```diff
@@ -0,0 +1,206 @@
+import sys
+
+import pytest
+
+if sys.version_info < (3, 10):
+  pytest.skip('Python 3.10 or higher is required', allow_module_level=True)
```
ParthSareen, line 6: Must be done for our older test matrix as it would fail on parsing the new `UnionType`s

ollama/_utils.py
```diff
@@ -0,0 +1,213 @@
...
+      if len(non_none_args) == 1:
+        return _get_json_type(non_none_args[0])
+      # For multiple types (e.g., int | str | None), return array of types
+      return [_get_json_type(arg) for arg in non_none_args]
+    return 'null'
+
+  # Handle generic types (List[int], Dict[str, int], etc.)
+  if get_origin(python_type) is not None:
+    # Get the base type (List, Dict, etc.)
+    base_type = TYPE_MAP.get(get_origin(python_type), None)
```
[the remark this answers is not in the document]
ParthSareen: Given the structure of the JSON we need:
eg: 
```JSON
{
    "type": "function",
    "function": {
        "name": "add_numbers",
        "description": "Add two numbers together.",
        "parameters": {
            "type": "object",
            "required": [
                "x"
            ],
            "properties": {
                "x": {
                    "type": "integer",
                    "description": "The first number"
                },
                "y": {
                    "type": "integer",
                    "description": "The second number"
                }
            }
        },
        "return_type": "integer"
    }
}
```
The `parameters` key has a type of `JsonSchemaValue` defined in `_types.py` where pydantic doesn't do any serialization on the string values which should be mapped to valid JSON. So this means if it sees `typing.List` as the type in python it leaves it as is rather than mapping it to `array`.

ollama/_utils.py
```diff
@@ -0,0 +1,213 @@
...
+}
+
+
+def _get_json_type(python_type: Any) -> str | List[str]:
+  # Handle Optional types (Union[type, None] and type | None)
+  if is_union(python_type):
+    args = get_args(python_type)
+    # Filter out None/NoneType from union args
+    non_none_args = [arg for arg in args if arg not in (None, type(None))]
+    if non_none_args:
```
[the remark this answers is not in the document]
ParthSareen: Very go like haha I'll make the change 🦙

ollama/_types.py
```diff
@@ -229,12 +230,14 @@ class Function(SubscriptableBaseModel):
...
       required: Optional[Sequence[str]] = None
       properties: Optional[JsonSchemaValue] = None
 
     parameters: Parameters
 
+    return_type: Optional[Union[str, List[str]]] = None
```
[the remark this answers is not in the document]
ParthSareen: It is not but is built into other libs - thought we should have it exposed in case we need it later but happy to take out. Don't have a strong opinion here

ParthSareen:
Blocked on https://github.com/ollama/ollama/pull/7613 to support typing options

ollama/_utils.py
```diff
@@ -0,0 +1,212 @@
...
+def _get_json_type(python_type: Any) -> str | List[str]:
+  # Handle Optional types (Union[type, None] and type | None)
+  if is_union(python_type):
+    args = get_args(python_type)
+    # Filter out None/NoneType from union args
+    if non_none_args := [arg for arg in args if arg not in (None, type(None))]:
+      if len(non_none_args) == 1:
+        return _get_json_type(non_none_args[0])
+      # For multiple return types (e.g., int | str | None), return stringified array of types -> "['integer', 'string', 'null']"
+      return str([_get_json_type(arg) for arg in non_none_args]).replace(' ', '')
```
ParthSareen, line 70: added a stringify here to not have to change the types on ollama/ollama repo. Function calling working as expected with change 😁

ollama/_json_type_map.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import Dict, Set as TypeSet, TypeVar
+
+T = TypeVar('T')
+if sys.version_info >= (3, 10):
+  from types import UnionType
+else:
+  UnionType = Union[T]
+
+# Python doesn't have a type serializer, so we need to map types to JSON types
+TYPE_MAP = {
```
ParthSareen, line 13: Need this ugly thing to make sure serialization is done correctly from type or str -> str

ollama/_types.py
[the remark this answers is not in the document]
ParthSareen: Decided to rm this. We can easily bake in anytime if needed

ollama/_types.py
```diff
@@ -229,9 +225,19 @@ class Function(SubscriptableBaseModel):
     description: str
 
     class Parameters(SubscriptableBaseModel):
-      type: str
+      type: Literal['object'] = 'object'
       required: Optional[Sequence[str]] = None
-      properties: Optional[JsonSchemaValue] = None
+
+      class Property(SubscriptableBaseModel):
+        model_config = ConfigDict(arbitrary_types_allowed=True)
```
ParthSareen, line 232: Callout: Needed if ever using custom types or `UnionType` serialization

ollama/_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+  if 'returns:' in lowered_doc_string:
+    # Return section can be captured and used
+    args_section = args_section.split('returns:')[0]
+
+  if 'yields:' in lowered_doc_string:
+    args_section = args_section.split('yields:')[0]
+
+  cur_var = None
+  for line in args_section.split('\n'):
```
mxyng, line 32: this can be significantly simpler if you start with this

```python
def parse(docstring):
  descriptions = defaultdict(list)

  key = '_'
  for line in docstring.splitlines():
    if line.lower().startswith('args:'):
      key = 'args'
    elif line.lower().startswith('raises:') or line.lower().startswith('returns:') or line.lower().startswith('yields:'):
      key = ''

    if key:
      descriptions[key].append(line.strip())

  return descriptions
```

ollama/_utils.py
```diff
@@ -0,0 +1,103 @@
...
+      '__annotations__': {k: v.annotation for k, v in inspect.signature(func).parameters.items()},
+      '__signature__': inspect.signature(func),
+      '__doc__': inspect.getdoc(func),
+    },
+  ).model_json_schema()
+
+  properties = {}
+  required = []
+  parsed_docstring = _parse_docstring(schema.get('description'))
+  for k, v in schema.get('properties', {}).items():
```
mxyng, line 71: You can replace the map value in this loop instead of creating temporary values:

```python
  for k, v in schema.get('properties', {}).items():
    schema['properties'][k] = {
      'type': ', '.join([i.get('type') for i in v.get('anyOf')]) if 'anyOf' in v else v.get('type'),
      'description': ' '.join(line for line in descriptions['args'] if line.startswith(f'{k} (')),
    }

    if 'null' in schema['properties'][k]['type'].split(', '):
      schema['required'].remove(k)
```

ollama/_client.py
```diff
@@ -1075,9 +1079,15 @@ def _copy_messages(messages: Optional[Sequence[Union[Mapping[str, Any], Message]
     )
 
 
-def _copy_tools(tools: Optional[Sequence[Union[Mapping[str, Any], Tool]]]) -> Iterator[Tool]:
-  for tool in tools or []:
-    yield Tool.model_validate(tool)
+def _copy_tools(tools: Optional[Sequence[Union[Mapping[str, Any], Tool, Callable]]] = None) -> Iterator[Tool]:
+  if not tools:
+    return []
```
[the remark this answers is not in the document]
ParthSareen: Shouldn't that be the case? It would mean that no tools get assigned on the request shown below (which is what is expected)

```py
        model=model,
        messages=[message for message in _copy_messages(messages)],
        tools=[tool for tool in _copy_tools(tools)],
        stream=stream,
 ```

https://github.com/ollama/ollama-python/pull/321/files/16c868a9375af412d5e4be21ef3060afa256faf6#diff-0f246a9c084dd2ef9d4a58c02fb818ac4a114c34877e558cac88ab351daae9eeR790-L788

File: ollama/_client.py [hunk 14]
```diff
@@ -293,6 +297,30 @@ def chat(
...
+      tools:
+        A JSON schema as a dict, an Ollama Tool or a Python Function.
+        Python functions need to follow Google style docstrings to be converted to an Ollama Tool.
+        For more information, see: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings
+      stream: Whether to stream the response.
+      format: The format of the response.
+
+    Example:
+      def add_two_numbers(a: int, b: int) -> int:
+        \"""
```
mxyng, line 310: You can change the top level quotes to `'''`

ollama/_client.py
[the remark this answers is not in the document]
mxyng: Sorry missed a word. Meant to say "it shouldn't be _necessary_"

ollama/_client.py
```diff
@@ -1152,6 +1207,8 @@ def _parse_host(host: Optional[str]) -> str:
   'https://[0001:002:003:0004::1]:56789/path'
   >>> _parse_host('[0001:002:003:0004::1]:56789/path/')
   'http://[0001:002:003:0004::1]:56789/path'
+  >>> _parse_host('http://host.docker.internal:11434/path')
+  'http://host.docker.internal:11434/path'
```
mxyng, line 1211: Why the new new test case? This is equivalent to `https://example.com:56789/path` above

ParthSareen: Oops thanks for the catch - did not mean for this one to be in

ollama/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+def _parse_docstring(doc_string: Union[str, None]) -> dict[str, str]:
+  parsed_docstring = defaultdict(str)
+  if not doc_string:
+    return parsed_docstring
+
+  lowered_doc_string = doc_string.lower()
+
+  key = hash(doc_string)
+  parsed_docstring[key] = ''
```
mxyng, line 18: this shouldn't be necessary with the defaultdict

ollama/_utils.py
```diff
@@ -0,0 +1,81 @@
...
+import pydantic
+from ollama._types import Tool
+
+
+def _parse_docstring(doc_string: Union[str, None]) -> dict[str, str]:
+  parsed_docstring = defaultdict(str)
+  if not doc_string:
+    return parsed_docstring
+
+  lowered_doc_string = doc_string.lower()
```
mxyng, line 15: What's the objective with lowering the entire docstring? AFAICT you only need specific values like `Args:` and `Returns:` etc. lowered

tests/test_client.py
```diff
@@ -982,3 +983,67 @@ def test_headers():
...
+      'description': 'Test function',
+      'parameters': {
+        'type': 'object',
+        'properties': {'x': {'type': 'string', 'description': 'A string'}},
+        'required': ['x'],
+      },
+    },
+  }
+
+  tool_json = json.loads(json.dumps(tool_dict))
```
mxyng, line 1028: What's the point of this? Can't you use `tool_dict` directly? `tool_json` should be the same as `tool_dict`

ollama/_utils.py
ParthSareen: Initially thought that if we were calling `.lower()` on the same thing multiple times it would be better do just lower it once. Don't think I ended actually needing it on the whole docstring so will probably change it.

tests/test_client.py
```diff
@@ -982,3 +983,67 @@ def test_headers():
...
+  # Test that malformed tool dictionaries are rejected
+  # Raises ValidationError when used as it is a generator
+  with pytest.raises(ValidationError):
+    invalid_tool = {'type': 'invalid_type', 'function': {'name': 'test'}}
+    list(_copy_tools([invalid_tool]))
+
+  # Test missing required fields
+  incomplete_tool = {
+    'type': 'function',
+    'function': {'name': 'test'},  # missing description and parameters
```
mxyng, line 1046: Is parameters required?

tests/test_type_serialization.py
```diff
@@ -13,3 +16,81 @@ def test_image_serialization():
   b64_str = 'dGVzdCBiYXNlNjQgc3RyaW5n'
   img = Image(value=b64_str)
   assert img.model_dump() == b64decode(b64_str).decode()
+
+
+def test_tool_serialization():
```
mxyng, line 21: Is this still useful? It's only testing pydantic serialization now AFAIK

ParthSareen: Agree - should rm

tests/test_client.py
ParthSareen: I was messing around with some stuff haha snuck in!

tests/test_client.py
ParthSareen: So in the previous behavior - no.
But since the last Pydantic PR you have to provide Parameters. Tested out both the OpenAI Python SDK with Ollama and the current published version of ollama-python. I think it might be worthwhile to have the Tool class have optional on all params to make sure it is backwards compatible. 

Seems like function calling still works with not a fully-fledged JSON (I'm sure it's flakey but not our job).

Proposed Tool def'n:
```py
class Tool(SubscriptableBaseModel):
  type: Optional[Literal['function']] = 'function'

  class Function(SubscriptableBaseModel):
    name: Optional[str] = None
    description: Optional[str] = None

    class Parameters(SubscriptableBaseModel):
      type: Optional[Literal['object']] = 'object'
      required: Optional[Sequence[str]] = None

      class Property(SubscriptableBaseModel):
        model_config = ConfigDict(arbitrary_types_allowed=True)

        type: Optional[str] = None
        description: Optional[str] = None

      properties: Optional[Mapping[str, Property]] = None

    parameters: Optional[Parameters] = None

  function: Optional[Function] = None
```

ollama/_client.py
```diff
@@ -771,6 +799,30 @@ async def chat(
...
+        Args:
+          a: First number to add
+          b: Second number to add
+
+        Returns:
+          int: The sum of a and b
+        '''
+        return a + b
+
+      client.chat(model='llama3.1:8b', tools=[add_two_numbers], messages=[...])
```
mxyng, line 824: minor: this should `await`

ollama/_utils.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def _parse_docstring(doc_string: Union[str, None]) -> dict[str, str]:
+  parsed_docstring = defaultdict(str)
+  if not doc_string:
+    return parsed_docstring
+
+  key = hash(doc_string)
+  for line in doc_string.splitlines():
+    lowered_line = line.lower()
```
mxyng, line 17: might be a good idea to `strip` here so it'll match trailing whitespace on `args:`

ParthSareen: Yes smart!

ollama/_utils.py
```diff
@@ -0,0 +1,79 @@
...
+      key = '_'
+
+    else:
+      # maybe change to a list and join later
+      parsed_docstring[key] += f'{line.strip()}\n'
+
+  last_key = None
+  for line in parsed_docstring['args'].splitlines():
+    line = line.strip()
+    if ':' in line and not line.lower().startswith('args:'):
```
mxyng, line 30: `args:` should not be in the line anymore so the second conditional can be omitted

ollama/_utils.py
```diff
@@ -0,0 +1,79 @@
...
+      # Split on first occurrence of '(' or ':' to separate arg name from description
+      split_char = '(' if '(' in line else ':'
```
mxyng, line 32: what happens if there's a `(` in the description body?, e.g.

```
Args:
  param: do thing (default: something)
```

ParthSareen: This was a really good point. Added some regex to have more consistent parsing

ParthSareen: Added test cases as well

ollama/_utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+  last_key = None
+  for line in parsed_docstring['args'].splitlines():
+    line = line.strip()
+    if ':' in line:
+      # Split the line on either:
+      # 1. A parenthetical expression like (integer) - captured in group 1
+      # 2. A colon :
+      # Followed by optional whitespace. Only split on first occurrence.
+      parts = re.split(r'(?:\(([^)]*)\)|:)\s*', line, maxsplit=1)
```
mxyng, line 36: I tend to avoid regexp when possible since it's hard to grok. In this scenario, a simpler solution would be to split on the mandatory `:` than parse the pre-colon and post-colon sections independently. Here's an example that passes your tests

```python
  for line in parsed_docstring['args'].splitlines():
    pre, _, post = line.partition(':')
    if not pre.strip():
      continue
    if not post.strip() and last_key:
      parsed_docstring[last_key] += ' ' + pre
      continue

    arg_name, _, _ = pre.replace('(', ' ').partition(' ')
    last_key = arg_name.strip()

    parsed_docstring[last_key] = post.strip()
```

ollama/_utils.py
```diff
@@ -0,0 +1,87 @@
...
+      '__doc__': parsed_docstring[doc_string_hash],
+    },
+  ).model_json_schema()
+
+  for k, v in schema.get('properties', {}).items():
+    # If type is missing, the default is string
+    types = {t.get('type', 'string') for t in v.get('anyOf')} if 'anyOf' in v else {v.get('type', 'string')}
+    if 'null' in types:
+      schema['required'].remove(k)
+      types.discard('null')
```
mxyng, line 72: If `null` is the only type (for some reason), this will be an empty string

ollama/_utils.py
ParthSareen: TODO @ParthSareen to spin out issue

mxyng (approved):

20246688:
I found that the ollama-Python/examples/tools.py script cannot specify an IP address.

PR #327, opened 2024-11-20
Bugfix/image encoding

ParthSareen:
Images are not being encoded correctly given a b64 string - it tries to encode it twice.

Updated the types for Image as well as the serialization to better handle these cases - added tests as well

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -97,6 +97,10 @@ class Image(BaseModel):
     if isinstance(self.value, str):
       if Path(self.value).exists():
         return b64encode(Path(self.value).read_bytes()).decode()
+
+      if self.value.split('.')[-1] in ('png', 'jpg', 'jpeg', 'webp'):
+        raise ValueError(f'File {self.value} does not exist')
+
       try:
         # Try to decode to check if it's already base64
         b64decode(self.value)
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -102,7 +102,7 @@ class Image(BaseModel):
         b64decode(self.value)
         return self.value
       except Exception:
-        return self.value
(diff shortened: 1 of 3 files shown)


ollama/_types.py
```diff
@@ -89,16 +87,26 @@ class BaseGenerateRequest(BaseStreamableRequest):
...
-    return self.value
+    if isinstance(self.value, (Path, bytes)):
+      return b64encode(self.value.read_bytes() if isinstance(self.value, Path) else self.value).decode()
+
+    if isinstance(self.value, str):
+      if Path(self.value).exists():
+        return b64encode(Path(self.value).read_bytes()).decode()
+
+      if self.value.split('.')[-1] in ('png', 'jpg', 'jpeg', 'webp'):
+        raise ValueError(f'File {self.value} does not exist')
```
ParthSareen, line 102: This is to handle the edge case where a path does not exist but could be a valid b64 string - would lead to weird behavior

tests/test_type_serialization.py
```diff
@@ -1,16 +1,48 @@
...
+def test_image_serialization_base64_string():
   b64_str = 'dGVzdCBiYXNlNjQgc3RyaW5n'
   img = Image(value=b64_str)
-  assert img.model_dump() == b64decode(b64_str).decode()
+  assert img.model_dump() == b64_str  # Should return as-is if valid base64
+
+
+def test_image_serialization_plain_string():
+  img = Image(value='not a path or base64')
+  assert img.model_dump() == 'not a path or base64'  # Should return as-is
```
ParthSareen, line 24: Currently just a passthrough. Since spaces are valid b64 I'm hesitant to split on it and check the length. although we could do a heuristic to catch it greedily

mxyng (approved):

PR #334, opened 2024-11-24
fix: hasattr checks if attr is None

mxyng:
```
>>> import ollama
>>> r = ollama.chat(model='tmp')
>>> r.message
Message(role='assistant', content='', images=None, tool_calls=None)
>>> 'tool_calls' in r.message
False
>>> 'role' in r.message
True
```

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -45,6 +45,9 @@ class SubscriptableBaseModel(BaseModel):
     >>> msg['tool_calls'] = [Message.ToolCall(function=Message.ToolCall.Function(name='foo', arguments={}))]
     >>> 'tool_calls' in msg
     True
+    >>> msg['tool_calls'] = None
+    >>> 'tool_calls' in msg
+    True
     >>> tool = Tool()
     >>> 'type' in tool
     True
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -45,8 +45,17 @@ class SubscriptableBaseModel(BaseModel):
     >>> msg['tool_calls'] = [Message.ToolCall(function=Message.ToolCall.Function(name='foo', arguments={}))]
     >>> 'tool_calls' in msg
     True
+    >>> tool = Tool()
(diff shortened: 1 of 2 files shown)


jmorganca (approved):

jmorganca (approved):

ollama/_types.py
```diff
@@ -23,7 +23,30 @@ def __setitem__(self, key: str, value: Any) -> None:
...
+    >>> 'tool_calls' in msg
+    False
+    >>> msg['tool_calls'] = []
+    >>> 'tool_calls' in msg
+    True
+    >>> msg['tool_calls'] = [Message.ToolCall(function=Message.ToolCall.Function(name='foo', arguments={}))]
+    >>> 'tool_calls' in msg
+    True
+    """
+    return key in self.model_fields_set
```
jmorganca, line 49: Will this inadvertently filter out fields with a default value?

mxyng: I don't think so

> Returns the set of fields that have been explicitly set on this model instance.

jmorganca: Did we want this?

```
return key in self.model_fields_set or key is not None
```

I don't think default values are in `model_fields_set`

```
@property
    def model_fields_set(self) -> set[str]:
        """Returns the set of fields that have been explicitly set on this model instance.

        Returns:
            A set of strings representing the fields that have been set,
                i.e. that were not filled from defaults.
        """
        return self.__pydantic_fields_set__
```

ParthSareen: Screenshot 2024-11-23 at 6 03 36 PM
Just checked - default values are omitted in this case which is not what we want.

jmorganca, line 49:
```suggestion
    if key in self.model_fields_set:
        return True

    if key in self.model_fields:
        return self.model_fields[key].default is not None

    return False
```

ParthSareen (approved):

PR #337, opened 2024-11-24
Change proposed in example/chat-with-history.py to overcome runtime error.

shahil-yadav:
append() method accepts only 1 argument. We can use extend() method as an alternative.

### DIFF
--- a/examples/chat-with-history.py
+++ b/examples/chat-with-history.py
@@ -31,8 +31,8 @@ while True:
   )
 
   # Add the response to the messages to maintain the history
-  messages.append(
-    {'role': 'user', 'content': user_input},
-    {'role': 'assistant', 'content': response.message.content},
+  messages.extend(
+        [{'role': 'user', 'content': user_input},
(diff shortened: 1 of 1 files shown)


examples/chat-with-history.py
```diff
@@ -31,8 +31,8 @@
   )
 
   # Add the response to the messages to maintain the history
-  messages.append(
-    {'role': 'user', 'content': user_input},
-    {'role': 'assistant', 'content': response.message.content},
+  messages.extend(
+        [{'role': 'user', 'content': user_input},
+         {'role': 'assistant', 'content': response.message.content}]
   )
```
ParthSareen, line 37:
```suggestion
  messages += [
    {'role': 'user', 'content': user_input},
    {'role': 'assistant', 'content': response.message.content},
  ]
```

ParthSareen: I think this is a bit cleaner - thanks for the catch though!

PR #344, opened 2024-11-27
Fix broken `Client.chat` overload type annotations

pythongirl325:
The overload type annotations for `Client.chat` did not reflect the new tool function API.


ParthSareen (approved):
Seems like I forgot to update throughout the signatures. Thanks for the catch! 😄

hallettj:
I'm hitting type errors from pyright as a result of this change when calling `Client.chat` even though I'm not passing in a `tools` argument:

```
error: Type of "chat" is partially unknown
    Type of "chat" is "Overload[(model: str = '', messages: Sequence[Mapping[str, Any] | Message] | None = None, *, tools: Sequence[Mapping[str, Any] | Tool | ((...) -> Unknown)] | None = None, stream: Literal[False] = False, format: Dict[str, Any] | Literal['', 'json'] | None = None, options: Mapping[str, Any] | Options | None = None, keep_alive: float | str | None = None) -> ChatResponse, (model: str = '', messages: Sequence[Mapping[str, Any] | Message] | None = None, *, tools: Sequence[Mapping[str, Any] | Tool | ((...) -> Unknown)] | None = None, stream: Literal[True] = True, format: Dict[str, Any] | Literal['', 'json'] | None = None, options: Mapping[str, Any] | Options | None = None, keep_alive: float | str | None = None) -> Iterator[ChatResponse]]" (reportUnknownMemberType)
```

The call site looks like this:

```py
response = self.client.chat(
    model=self.model,  # str
    messages=messages, # list[Message]
    stream=False,      # bool
    options=options,   # Options
)
```

The problem disappears if I edit the signatures of `Client.chat` to remove `Callable` from the tools parameter union.

I'm using ollama v0.4.7 and pyright v1.1.398.

Most likely I'm doing something to trigger the error, but after lots of troubleshooting I have no idea what I can do to fix the error other than downgrading to ollama v0.4.1.

ParthSareen:
Hey @hallettj, sorry about that :/ 

Not sure why pyright is complaining since it's init to None. Will check it out a bit more to see what's up.

hallettj:
@ParthSareen Thanks for following up! I'll add that I can resolve the error by editing the signatures for `Client.chat` to specify type arguments to `Callable`. If I replace each instance of `Callable` with `Callable[..., Any]` that seems to satisfy pyright.

ParthSareen:
@hallettj that makes me think that it's defaulting to a `Callable` from pyright's perspective even though it should be a `None` it's probably something small or a discrepancy with pyright. Thanks for letting me know!

hallettj:
@ParthSareen I don't think the error has anything to do with the call site. I think the issue is that pyright is not able to fully process the method signatures. And it looks to me like the issue is that pyright can't resolve `Callable` without arguments to a type. Whether or not you call `chat` with a `tools` argument, pyright tries to fully resolve every type in every union for every parameter of the method, and it has a problem with `Callable`.

PR #354, opened 2024-12-01
Structured outputs support with examples

ParthSareen:
Enable structured outputs support https://github.com/ollama/ollama/pull/7900


ParthSareen:
Adding generate updates here as well!

ollama/_client.py
```diff
@@ -186,7 +188,7 @@ def generate(
     context: Optional[Sequence[int]] = None,
     stream: Literal[False] = False,
     raw: bool = False,
-    format: Optional[Literal['', 'json']] = None,
+    format: Optional[Union[Literal['', 'json'], Type[BaseModel]]] = None,
```
mxyng, line 191: This should also accept arbitrary Python types by way of `pydantic.TypeAdapter` as well as the `TypeAdapter` itself

mxyng: Never mind

examples/structured-outputs.py
```diff
@@ -0,0 +1,56 @@
...
+
+## Serialize the dataclass schema using asdict
+response = chat(
+  model='llama3.1:8b',
+  messages=[{'role': 'user', 'content': 'I have two friends. The first is Ollama 22 years old busy saving the world, and the second is Alonso 23 years old and wants to hang out. Return a list of friends in JSON format'}],
+  format=friends_list_schema,
+  options={'temperature': 0},
+)
+
+friends_response = FriendList(friends=[FriendInfo(**friend) for friend in json.loads(response.message.content)['friends']])
```
ParthSareen, line 55: this is really ugly but from what i understand the only way to serialize into a nested object correctly

ParthSareen: okay im gonna get rid of this i dont like it LOL

mxyng (approved):

PR #363, opened 2024-12-06
Update async-tools.py with multiple tools calls

damiencorpataux:
Hey there !

I wanted to play with the basics of multiple tools call, here is an update of the example that works for me.

Thank you & Cheers !

### DIFF
--- a/examples/async-tools.py
+++ b/examples/async-tools.py
@@ -14,14 +14,14 @@ def add_two_numbers(a: int, b: int) -> int:
   Returns:
     int: The sum of the two numbers
   """
-  return a + b
+  return float(a) + float(b)
 
 
 def subtract_two_numbers(a: int, b: int) -> int:
   """
   Subtract two numbers
   """
-  return a - b
+  return float(a) - float(b)
 
 
 # Tools can still be manually defined and passed into chat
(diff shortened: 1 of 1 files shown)


damiencorpataux:
Just added casting of the arguments in tools functions.

I ran the updated example again after posting my PR and got an error: `TypeError: unsupported operand type(s) for -: 'str' and 'str'` because `tool.function.arguments` suddenly contained args of type `str`:
```
Prompt: What is three plus one minus two ?
Calling function: subtract_two_numbers
Arguments: {'a': '3', 'b': '1'}
...(TypeError)...
```
instead of `int` (as of my previous runs):
```
Prompt: What is three plus one minus two ?
Calling function: subtract_two_numbers
Arguments: {'a': 3, 'b': 2}
Function called: {'function': 'subtract_two_numbers', 'output': 1}
Final response: To solve the problem, first add 3 and 1, then subtract 2.
```

Here is a cast of the error:
https://asciinema.org/a/Egg7wxg0qEqteWd4DPG4ItB0H

ParthSareen:
Hey @damiencorpataux! Appreciate you PR'ing this. This is a pretty common problem with smaller models as the JSON they produce isn't always consistent with the types provided. This doesn't need to be codified into the example but at some point we'll have structured generation of tools which should get rid of this problem. Hope this helps!

Issue #366, opened 2024-12-07
Breaking Change for `langchain-ollama==0.2.1` on `ollama==0.4.3`


ross-harrison-fox:
Receive Validation code when running the bellow snippet with the following package combination:

* `langchain==0.3.10`
* `ollama==0.4.3`
* `langchain-ollama==0.2.1`

```python
# get embedding and LLM models
from langchain_ollama import OllamaEmbeddings, OllamaLLM

embeddings = OllamaEmbeddings(model="llama3", base_url="http://localhost:11434")
llm = OllamaLLM(model="llama3", format='', base_url="http://localhost:11434", verbose=True)

prompt = "Hello There"
response = llm.invoke(prompt)
```

Expected Behavior: 

Some response

Actual:

See validation error

```
19 prompt = "Hello There"
---> 21 response = llm.invoke(prompt)

File ~\miniconda3\Lib\site-packages\langchain_core\language_models\chat_models.py:286, in BaseChatModel.invoke(self, input, config, stop, **kwargs)
    275 def invoke(
    276     self,
    277     input: LanguageModelInput,
   (...)
    281     **kwargs: Any,
    282 ) -> BaseMessage:
    283     config = ensure_config(config)
    284     return cast(
    285         ChatGeneration,
--> 286         self.generate_prompt(
    287             [self._convert_input(input)],
    288             stop=stop,
    289             callbacks=config.get("callbacks"),
    290             tags=config.get("tags"),
    291             metadata=config.get("metadata"),
    292             run_name=config.get("run_name"),
    293             run_id=config.pop("run_id", None),
    294             **kwargs,
    295         ).generations[0][0],
    296     ).message

File ~\miniconda3\Lib\site-packages\langchain_core\language_models\chat_models.py:786, in BaseChatModel.generate_prompt(self, prompts, stop, callbacks, **kwargs)
    778 def generate_prompt(
    779     self,
    780     prompts: list[PromptValue],
   (...)
    783     **kwargs: Any,
    784 ) -> LLMResult:
    785     prompt_messages = [p.to_messages() for p in prompts]
--> 786     return self.generate(prompt_messages, stop=stop, callbacks=callbacks, **kwargs)

File ~\miniconda3\Lib\site-packages\langchain_core\language_models\chat_models.py:643, in BaseChatModel.generate(self, messages, stop, callbacks, tags, metadata, run_name, run_id, **kwargs)
    641         if run_managers:
[...]

PR #368, opened 2024-12-08
fix: address pydantic validation errors during ollama LLM invoke requests

nadeemcite:
This PR resolves a pydantic validation issue encountered when invoking Ollama’s LLM generation requests, and addresses issue #366 

This PR resolves a pydantic validation issue encountered when invoking Ollama’s LLM generation requests. Previously, the format field for GenerateRequest strictly required a literal value of 'json'. When the format field was omitted or passed as an empty string, Pydantic validation failed, producing errors similar to the following:

```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for GenerateRequest
format.literal['json']
  Input should be 'json' [type=literal_error, input_value='', input_type=str]
  
format.dict[str,any]
  Input should be a valid dictionary [type=dict_type, input_value='', input_type=str]
```

By expanding the format union type to include both the empty string '' and 'json' values, the request validation now passes even when format is not explicitly defined or is empty. This change ensures that developers have more flexibility in specifying their request parameters and that the underlying validation does not fail in cases where format is absent or intentionally left blank.

### Key Changes:
* Updated the format field from Optional[Union[Literal['json'], JsonSchemaValue]] to Optional[Union[Literal['', 'json'], JsonSchemaValue]].
* Ensured that requests with an empty format do not cause validation errors.

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -151,7 +151,7 @@ class BaseGenerateRequest(BaseStreamableRequest):
   options: Optional[Union[Mapping[str, Any], Options]] = None
   'Options to use for the request.'
 
-  format: Optional[Union[Literal['json'], JsonSchemaValue]] = None
+  format: Optional[Union[Literal['', 'json'], JsonSchemaValue]] = None
   'Format of the response.'
 
   keep_alive: Optional[Union[float, str]] = None


varun-d:
Thank you for raising this pull request

jmorganca:
Thanks so much for the PR! I've fixed it in a few other places as well in https://github.com/ollama/ollama-python/pull/369 for expediency.

PR #370, opened 2024-12-09
Fix AttributeError in structured-outputs example by adjusting response parsing

tdiprima:
### Description
This PR addresses an issue where the structured-outputs example script throws an `AttributeError` when accessing the `message` field of the `response` object. The root cause is that the `response` object is a dictionary, not an object with attributes.

### Changes Made
1. Replaced `response.message.content` with `response["message"]["content"]` to correctly access the content.
2. Added a raw response printout (`print("Raw Response:", response)`) to help debug unexpected response structures.
3. Included error handling to check for the existence of `message` and `content` keys before attempting to parse the response.
4. Enhanced the readability and formatting of the prompt sent to the `chat` function.

### Code Changes
```python
# Fixed response parsing
if "message" in response and "content" in response["message"]:
    friends_response = FriendList.model_validate_json(response["message"]["content"])
    print(friends_response)
else:
    print("Unexpected response structure:", response)
```

### Steps to Reproduce the Error
1. Run the original `structured-outputs.py` script.
2. Observe the `AttributeError: 'dict' object has no attribute 'message'`.

### Steps to Verify the Fix
1. Replace the original script with the updated code in this PR.
2. Run the script.
3. Verify that the response is successfully parsed and printed without any errors.

### Testing
- Tested locally using the latest version of the `ollama` library.
- Added a print statement to ensure the response structure aligns with expectations.

### DIFF
--- a/examples/structured-outputs.py
+++ b/examples/structured-outputs.py
@@ -1,26 +1,37 @@
 from ollama import chat
 from pydantic import BaseModel
 
-
 # Define the schema for the response
 class FriendInfo(BaseModel):
-  name: str
-  age: int
-  is_available: bool
-
+    name: str
+    age: int
+    is_available: bool
 
 class FriendList(BaseModel):
-  friends: list[FriendInfo]
+    friends: list[FriendInfo]
 
-
-# schema = {'type': 'object', 'properties': {'friends': {'type': 'array', 'items': {'type': 'object', 'properties': {'name': {'type': 'string'}, 'age': {'type': 'integer'}, 'is_available': {'type': 'boolean'}}, 'required': ['name', 'age', 'is_available']}}}, 'required': ['friends']}
+# Send a request to the chat model
 response = chat(
-  model='llama3.1:8b',
-  messages=[{'role': 'user', 'content': 'I have two friends. The first is Ollama 22 years old busy saving the world, and the second is Alonso 23 years old and wants to hang out. Return a list of friends in JSON format'}],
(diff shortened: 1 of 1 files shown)


ParthSareen:
Hey! We shouldn't be getting back a dict as part of the response. What version of ollama-python and ollama are you on? 

`pip show ollama`

tdiprima:
@ParthSareen Hi!  Version: 0.3.3

ParthSareen:
> @ParthSareen Hi!  Version: 0.3.3

Ahhh I think that's where your issue is. We recently updated it so they'd be returned as objects but still be accessible as dicts. 

Could you try to upgrade to `0.4.4`?

Let me know if that fixes it! - good chunk of other examples also use the direct access from an object.

tdiprima:
Thanks for suggesting 0.4.4! That said, I think it’s better to update the code to work across different versions instead of depending on just one. A version-agnostic approach would make the code more flexible and easier for other developers to use without worrying about matching a specific version.

ParthSareen:
> Thanks for suggesting 0.4.4! That said, I think it’s better to update the code to work across different versions instead of depending on just one. A version-agnostic approach would make the code more flexible and easier for other developers to use without worrying about matching a specific version.
> 
> TL;DR: 0.4.4 did fix the original code, but mine works just as well for either version of ollama.

So the old examples do follow that - and if you were on an old version you'd be able to access data like you always have. However for the new examples, it's best to show the most updated practices which is why they're only with the access methods. Hope that makes sense! Do appreciate the contribution and digging into it 😄

PR #373, opened 2024-12-12
fix TypeError for async-structured-outputs.py.

fujitatomoya:


examples/async-structured-outputs.py
```diff
@@ -11,7 +12,7 @@ class FriendInfo(BaseModel):
 
 
 class FriendList(BaseModel):
-  friends: list[FriendInfo]
```
fujitatomoya, line 14: this generates the following error with python 3.8.

```bash
tomoyafujita@~/DVT/03_OSS/github.com/ollama/ollama-python >python ./examples/async-structured-outputs.py
Traceback (most recent call last):
  File "./examples/async-structured-outputs.py", line 14, in <module>
    class FriendList(BaseModel):
  File "./examples/async-structured-outputs.py", line 15, in FriendList
    friends: list[FriendInfo]
TypeError: 'type' object is not subscriptable
```

introduced by https://github.com/ollama/ollama-python/pull/354, see more details for https://peps.python.org/pep-0585/.

fujitatomoya:
@ParthSareen can you take a look at this? just a minor cosmetic fix.

fujitatomoya:
@ParthSareen friendly ping

ParthSareen:
Hi @fujitatomoya thanks for the ping. I did think a bit about this and my take is:
- Python 3.8 is very old
- I would like to have best practices in examples - which means things like using newer patterns and features that the language offers
- It makes sense to leave this in for new examples with typing

Happy to discuss this more but it makes sense to leave this in and have the user update their typing if needed

fujitatomoya:
@ParthSareen 

okay then maybe we should upgrade https://github.com/ollama/ollama-python/blob/4daf4afdb63d01ee0b094426e4abcf0f5af55193/poetry.lock#L8?

current requirement leads to this problem?

ParthSareen:
@fujitatomoya We do support `python3.8` onwards - it's more so that it's recommended to be on a newer version which is why the typing in the examples follows the new methodology. So someone can still use ollama with python3.8 with the old way of doing typing.

fujitatomoya:
okay, this is just a cosmetic fix, not end of the world! thanks!

i just found that following,

https://github.com/ollama/ollama-python/blob/4daf4afdb63d01ee0b094426e4abcf0f5af55193/examples/structured-outputs-image.py#L3

thought that consistency would be better.

ParthSareen:
hmm you're right I didn't realize we weren't doing the new style for that - going to update that :)

Appreciate you bringing this up though! 😄

fujitatomoya:
@ParthSareen 

> it's recommended to be on a newer version which is why the typing in the examples follows the new methodology.

if you want me to patch it above? let me know!

fujitatomoya:
yeah no worries. just an examples but that is likely where the new people start... if we want to recommend the newer version of python, maybe that should be addressed instead.

ParthSareen:
Yeah I agree - although my gut says newer users would be on a newer version of Python. The installs for 3.8 are pretty low too which means that it's not going to be top of mind.

As for the patch - Could you update `structured-ouputs-image.py` to use the new typing methods? Appreciate it!

fujitatomoya:
yeah can do! will ping you once that is ready.

PR #379, opened 2024-12-14
test against real Ollama running in tests

samuelcolvin:
Uses [`ollama-action`](https://github.com/pydantic/ollama-action) which I built for [`pydantic-ai`](https://github.com/pydantic/pydantic-ai) - happy to transfer `ollama-action` to the `ollama` organization if you'd like.

By the way, can I **STRONGLY** encourage you to move the codebase to use 4 spaces for indents, not 2:

4 spaces is the default and almost ubiquitous in Python ecosystem, the switch should be a matter of just configuring ruff and running it.

Using 2 spaces (trivial though it may seem) will put off any experienced Python developer from contributing - exactly the people I guess you want to contribute.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -34,3 +34,24 @@ jobs:
           name: pytest-results-${{ matrix.python-version }}
           path: junit/test-results-${{ matrix.python-version }}.xml
         if: ${{ always() }}
+
+  test-live:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - run: pipx install poetry
+      - uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+          cache: poetry
+
+      - run: poetry install --with=dev
+
+      - uses: pydantic/ollama-action@v3
+        with:
+          model: qwen2:0.5b
+
+      - run: poetry run pytest tests/test_live.py -v --durations=100
+        env:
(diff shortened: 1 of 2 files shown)


samuelcolvin:
@mxyng please could you kick off tests here.

akx:
Agreed on the 4 spaces (some of the examples in the readme are 4-spaced anyway, as I noticed via https://github.com/ollama/ollama-python/commit/f7467d8c490a4e7b15b5b2f370bb9b3ed0e1fc3c)...

ParthSareen:
Hey @samuelcolvin - thanks for the PR but I don't think it makes a lot of sense to have testing against a live endpoint these tests. We should have integration tests with mocks or a smoke test but that can live separately from a github runner.
Thanks!

samuelcolvin:
Sorry, but I really don't agree. There's absolutely no way to ensure the library works with ollama unless you actually test against it.

If you build a web app, do you write your unit tests against the actual database, or write mocks and smoke tests and hope for the best?

If this was thousands of lines of code or took 20 minutes to run, I might understand, but it takes 40 seconds once cached, and has a tiny diff.

ParthSareen:
Hi @samuelcolvin. Don't think we'll be supporting this for now. We maintain a pretty strict contract between Ollama and our SDKs and it's something we work to ensure is compatible. Appreciate the recommendation and advice :)

samuelcolvin:
Okay, just out of interest, what's the downside you're worried about?

PR #380, opened 2024-12-14
support a custom HTTPX client in `Client` and `AsyncClient`

samuelcolvin:
As well as allowing a custom httpx client, this also improves type safety in `Client` and `AsyncClient` - previously `self._client` implicitly had type of `None`.

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -74,48 +74,54 @@ from ollama._types import (
 T = TypeVar('T')
 
 
-class BaseClient:
+class Client:
+  @overload
   def __init__(
     self,
-    client,
     host: Optional[str] = None,
-    follow_redirects: bool = True,
+    *,
+    follow_redirects: bool | None = None,
     timeout: Any = None,
     headers: Optional[Mapping[str, str]] = None,
-    **kwargs,
+    **httpx_kwargs: Any,
+  ) -> None: ...
+
+  @overload
+  def __init__(
+    self,
+    host: Optional[str] = None,
+    *,
+    client: httpx.Client,
+    headers: Optional[Mapping[str, str]] = None,
+  ) -> None: ...
+
+  def __init__(
+    self,
+    host: Optional[str] = None,
+    *,
+    client: httpx.Client | None = None,
+    follow_redirects: bool | None = None,
+    timeout: Any = None,
+    headers: Optional[Mapping[str, str]] = None,
+    **httpx_kwargs: Any,
   ) -> None:
-    """
-    Creates a httpx client. Default parameters are the same as those defined in httpx
-    except for the following:
-    - `follow_redirects`: True
-    - `timeout`: None
-    `kwargs` are passed to the httpx client.
-    """
+    self._host = _parse_host(host or os.getenv('OLLAMA_HOST'))
+    self._request_headers = _get_headers(headers)
+    if client:
+      assert follow_redirects is None, 'Cannot provide both `client` and `follow_redirects`'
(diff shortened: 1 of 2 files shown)


samuelcolvin:
This is needed for `pydantic-ai` to use this library rather than the OpenAI compatibility layer, see https://github.com/pydantic/pydantic-ai/issues/242.

mxyng:
Thanks for the PR. Can you provide some details in the description of why this change is necessary for `pydantic-ai`?

samuelcolvin:
We would like to reuse a set of http connections (encapsulated in an HTTPX client) when creating Ollama clients, so it's as cheap as possible to create new clients.

Please could you at least kick off tests.

By the way, you can set github to run tests automatically, even for new contributors; this change makes it much more attractive to contribute to the library.

mxyng:
Hi can you fix the tests? ollama-python currently targets 3.8 so the union types `|` are not supported. Please use `typing.Union`

ollama/_client.py
```diff
@@ -74,48 +74,54 @@
 T = TypeVar('T')
 
 
-class BaseClient:
+class Client:
+  @overload
```
mxyng, line 78: These overload aren't necessary since there's no overlap between the client and non-client versions

ollama/_client.py
```diff
@@ -144,22 +152,25 @@ def _request(
...
     stream: bool = False,
     **kwargs,
   ) -> Union[T, Iterator[T]]: ...
 
   def _request(
     self,
     cls: Type[T],
-    *args,
+    method: str,
+    path: str,
```
mxyng, line 165:
```suggestion
    path: str,
    *,
```

ollama/_client.py
```diff
@@ -74,48 +74,54 @@
...
+  ) -> None: ...
 
-  def _request_raw(self, *args, **kwargs):
-    r = self._client.request(*args, **kwargs)
+  def __init__(
+    self,
+    host: Optional[str] = None,
+    *,
+    client: httpx.Client | None = None,
+    follow_redirects: bool | None = None,
```
mxyng, line 103:
```suggestion
    client: Optional[httpx.Client] = None,
    follow_redirects: Optional[bool] = None,
```

mxyng (changes requested):

ollama/_client.py
samuelcolvin: They are to maintain proper typing, the point is you can't pass `client` and `follow_redirects` or `timeout` together. These overloads mean doing so will give a typing error.

ollama/_client.py
samuelcolvin: fixed.

ollama/_client.py
samuelcolvin: fixed

samuelcolvin:
Fixed suggestions I think, by the way it would be easier for contributors and less work for you if you let tests run automatically.

bgeesaman:
Friendly request to please take another peek at this PR. I’m here as a user of pydantic-ai and would love to be able to leverage ollama structured outputs and pydantic-ai for super powered local agents.😀

Really appreciate the hard work from both projects, so thank you all in advance!

rgbkrk:
Likewise on Pydantic AI + Ollama. I was saddened to find out that streaming tool calls don't work with the OpenAI compat layer. Would be nice to use the Ollama bindings directly.

rgbkrk:
Created a follow up PR that tries to make the diff small in hopes we can get this in soon. https://github.com/ollama/ollama-python/pull/618

Issue #375, opened 2024-12-12
Validation error when providing a base64 string to `Image(value=...)` in `Message(images=...)`


daniel-iliesh:
I am encountering a validation error while trying to create a `Message` object with an attached image using the `ollama._types.Message` and `ollama._types.Image` classes. 
According to the documentation, the `Message` constructor’s images parameter should accept a `Sequence[Image]`, and the Image constructor should accept a `str`, `bytes`, or `Path`. 
I am passing a single base64-encoded image string directly to the `Image(value=...)` constructor, but I receive a validation error indicating that the input is being interpreted as an `Image` object rather than a raw `string/bytes/path`.

### What I did:

- Constructed a base64-encoded string of an image.
- Created an `Image` object with `Image(value=my_base64_string)`.
- Passed a list of such `Image` objects into `Message(images=...)`.

### Expected behavior:
The `Message` object should be successfully created with the provided `Image` objects, each initialized from a raw base64 string.

### Actual behavior:
A pydantic validation error is raised:
```
3 validation errors for Image
value.str
  Input should be a valid string ...
value.bytes
  Input should be a valid bytes ...
value.lax-or-strict[path_validator()] 
  Input is not a valid path ...
```
#### Full error:
```
value.str
  Input should be a valid string [type=string_type, input_value=Image(value='iVBORw0KGgoA...nRTIIAAAAASUVORK5CYII='), input_type=Image]
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
value.bytes
  Input should be a valid bytes [type=bytes_type, input_value=Image(value='iVBORw0KGgoA...nRTIIAAAAASUVORK5CYII='), input_type=Image]
    For further information visit https://errors.pydantic.dev/2.10/v/bytes_type
value.lax-or-strict[lax=union[json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]],function-after[path_validator(), str]],strict=json-or-python[json=function-after[path_validator(), str],python=is-instance[Path]]]
  Input is not a valid path for <class 'pathlib.Path'> [type=path_type, input_value=Image(value='iVBORw0KGgoA...nRTIIAAAAASUVORK5CYII='), input_type=Image]
```
The error suggests that the `Image` constructor is somehow receiving another `Image` object instead of a base64 string.

### Additional details:
- I verified that images.values() returns a base64 str.
- I am not double-wrapping the Image objects in my code.
- Using Message(images=[Image(value=my_base64_string)]) triggers the error.

[...]

PR #386, opened 2024-12-17
fix: ensure Image objects can be used directly

akx:
Without the fix in the 2nd commit, `_copy_messages()` would raise a Pydantic error as `Image`s are being wrapped as `Image(value=Image(...))`. 

Related to #384 which exposes the model for general use.

This PR also removes Pillow from testing dependencies; it's not worth it to pull in a dependency just to generate a static 1x1 PNG we already know.


ParthSareen:
Could also be related to #375 - will check this out

ParthSareen:
Hi @akx - should fix some interface mismatch stuff here too so going to work off your PR :) Thanks for your contribution 🫡

PR #390, opened 2024-12-21
bugfix: wrap images correctly

ParthSareen:
There is currently an issue with image serialization and passing in Images as part of the message directly. Works off #386 shoutout to @akx!
Closes:
- https://github.com/ollama/ollama-python/pull/386
- https://github.com/ollama/ollama-python/issues/375


ollama/_client.py
```diff
@@ -189,7 +189,7 @@ def generate(
     stream: Literal[False] = False,
     raw: bool = False,
     format: Optional[Union[Literal['', 'json'], JsonSchemaValue]] = None,
-    images: Optional[Sequence[Union[str, bytes]]] = None,
```
ParthSareen, line 192: Going to revert this change to keep PR scope small for now

jmorganca (approved):

Issue #319, opened 2024-11-09
ollama.generate() with multimodal llama3.2-vision does not pass images array when raw=True


rylativity:
I am using ollama.generate() to process invoice document images. I would like to set `raw=True` so I can specify my own full prompt without it being passed into the default template.

Here is what I would like to be able to do:
```
ollama.generate(model="llama3.2-vision",
   prompt="Describe the contents of the photo",
   raw=True,
   images=[img_bytes]
)
```

However, when I set `raw=True` and pass a prompt and image to the model as shown in the screenshot below, the images array does not appear to get passed to the LLM as evidenced by the model's hallucinated response when asked to describe the image. 
<img width="803" alt="Screenshot 2024-11-09 at 5 37 09 PM" src="https://github.com/user-attachments/assets/a25b54dd-3e00-4ee6-a7bd-7788a6a98247">

In other instances I've even seen it explicitly state "there is no image for me to describe", further suggesting that the images are not making it to the LLM.

The next screenshot (below) shows a response when I set `raw=False` (default arg value) where the model is clearly receiving the image and able to provide an appropriate, expected response.
<img width="803" alt="Screenshot 2024-11-09 at 5 38 13 PM" src="https://github.com/user-attachments/assets/95cf92c5-2b67-4ad8-b862-4b4405fef72b">

I know that additional templating and formatting is being done to the inputs when `raw=False` based on the model's Ollama configuration, but I would think that there should be a way to pass images to a multimodal model when passing the prompt through as raw. 

I also understand that there are peculiarities with multimodal models in Ollama, so Is this a known limitation, is it a bug, or are there additional steps required to pass images alongside a prompt when `raw=True`?

Thanks in advance for your help.

ParthSareen:
Might be better to use the chat interface instead in this case:
```python
import ollama

response = ollama.chat(
    model='llama3.2-vision',
    messages=[{
        'role': 'user',
        'content': 'What is in this image?',
        'images': ['image.jpg']
    }]
)

print(response)
```

Let me know if this helps! I will check it out on the generate end in the meantime :)

ParthSareen:
Dug through our API and the Python SDK - I'd recommend to use the `.chat` functionality even in this case.

[...]

PR #392, opened 2024-12-22
Add support for prepending images to raw prompts and examples

rylativity:
This PR is related to a PR opened against ollama server to address unexpected behavior when passing both `images` and `raw=True` to a generate() call. The issue is described here - https://github.com/ollama/ollama-python/issues/319, and the open PR against ollama server is here - https://github.com/ollama/ollama/pull/8209.

These changes implement the necessary updates to the python client and generate() methods to allow passing an optional bool `prepend_images_to_raw_prompt` keyword argument to the generate() call. As optional, the change is intended to preserve (and not break) existing users. See associated PR against ollama server linked above for details about how this kwarg will work.

This PR also includes a new example titled `multimodal-raw-generate.py` which shows both how to use the proposed `prepend_images_to_raw_prompt` kwarg as well as how to manually place image tags within a raw prompt without the need for `prepend_images_to_raw_prompt`.


ParthSareen:
Hey @rylativity,

Thanks for the PR. I've thought a good bit about this and I think that for 99% of the cases, chat just makes more sense as you can easily transform structured JSON data. As you've already found, swapping out the image tag works for your use case but this is something that can possibly be inconsistent across models and possibly change in the near future. For that reason, going to have to close this one out for now.

Really appreciate you going deep into this though, and happy that you found something which fits your use case!

rylativity:
@ParthSareen in the associated PR to ollama server, I used the same image tags that ollama server already uses when handling images in chat mode - [see here](https://github.com/rylativity/ollama/blob/raw-prompt-images/server/routes.go#L263). Regarding your concern about models using different image tags in the future, these PRs don't add any overhead from that perspective, because the Chat image tags are equally susceptible to those potential changes and would need to be changed anyway. 

This is a different issue, but I might propose a tag lookup based on the model ID being used (all current models would map to `<image>` and others might raise an exception until implemented). 

Regarding the issue I opened the PRs against, I would urge you to consider at least approving the changes to the documentation showing how to pass those image tags in a raw prompt _**and**_ implementing some logic to raise an exception if some unsupported action with raw prompts and images is attempted. 

Right now, the behavior is not documented and will fail silently, which isn't good, but I am open to your opinions and happy to take a stab at implement the proposed changes above or any others that you suggest.

ParthSareen:
Sorry - what I meant to get at was that the responsibility of managing changing tags would be on the server end, managed by us and that's what I'd prefer. 

I think since this is a **raw** mode, it makes sense to not have an extra param and rather allow the user to pass in whatever they want. You are right about having this documented as even though this is a niche use case, it can be useful for certain folks. 
Before we add to the documentation, I would also prefer if we could talk a bit more about what the usage and error handling looks like. Appreciate your PR and convo here!

PR #398, opened 2024-12-29
client: improve error messaging on connection failure

ParthSareen:
Error messaging is unclear - especially on connection error
Also exposed the status code coming down but I'm not sure if we want that behavior

- [x] Tests

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -106,6 +106,9 @@ class BaseClient:
     )
 
 
+CONNECTION_ERROR_MESSAGE = 'Failed to connect to Ollama. Please check that Ollama is downloaded, running and accessible. https://ollama.com/download'
+
+
 class Client(BaseClient):
   def __init__(self, host: Optional[str] = None, **kwargs) -> None:
     super().__init__(httpx.Client, host, **kwargs)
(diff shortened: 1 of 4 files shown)


ollama/_types.py
```diff
@@ -524,3 +524,6 @@ def __init__(self, error: str, status_code: int = -1):
 
     self.status_code = status_code
     'HTTP status code of the response.'
+
+  def __str__(self) -> str:
+    return f'{self.error} (status code: {self.status_code})'
```
ParthSareen, line 529: status code does not get printed currently - this exposes it

ollama/_client.py
```diff
@@ -617,12 +619,14 @@ def __init__(self, host: Optional[str] = None, **kwargs) -> None:
...
-    r = await self._client.request(*args, **kwargs)
     try:
+      r = await self._client.request(*args, **kwargs)
       r.raise_for_status()
+      return r
     except httpx.HTTPStatusError as e:
       raise ResponseError(e.response.text, e.response.status_code) from None
-    return r
+    except httpx.ConnectError:
+      raise ResponseError('Failed to connect to Ollama. Please check that Ollama is downloaded, running and accessible. https://ollama.com/download', 503) from None
```
jmorganca, line 629: Shouldn't a 503 code be provided by the server? https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503 I wouldn't introduce any codes that aren't actually served by the API

ParthSareen: Yeah I'm debating if we should not have this as a response error since it's not even a response that is failing it's the connection from httpx itself

ParthSareen: Found a `ConnectionError` throwable which I think makes more sense in this setting

mxyng (approved):

ollama/_types.py
ParthSareen: No code, just the message from the server: 
image

ParthSareen: vs with the change: 
image

PR #400, opened 2024-12-30
fix: don't run lints on all Python versions

akx:
No need to spend time running the same lints 6 times.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -20,17 +20,27 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: poetry
       - run: poetry install --with=dev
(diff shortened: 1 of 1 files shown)


ParthSareen:
Hey! Thanks for this but we do need the tests running on the various versions as we still support Python3.8+ and there are breaking changes between the versions which we need to be cognizant of.

akx:
@ParthSareen _tests_ would still be run on every version, just not the static lints.

ParthSareen:
> @ParthSareen _tests_ would still be run on every version, just not the static lints. 

Missed the `lint:` tag initially - will give it a proper look later this week. Thanks!!

akx:
@ParthSareen Rebased for your convenience.

PR #403, opened 2025-01-01
Update the tools example to make it work

anilmuppalla:
this PR ensures the example can run as-is.

### DIFF
--- a/examples/tools.py
+++ b/examples/tools.py
@@ -1,5 +1,4 @@
-from ollama import chat
-from ollama import ChatResponse
+from ollama import ChatResponse, chat
 
 
 def add_two_numbers(a: int, b: int) -> int:
@@ -47,11 +46,12 @@ available_functions = {
   'add_two_numbers': add_two_numbers,
   'subtract_two_numbers': subtract_two_numbers,
 }
+model = 'llama3.2'
 
 response: ChatResponse = chat(
-  'llama3.1',
+  model=model,
   messages=messages,
-  tools=[add_two_numbers, subtract_two_numbers_tool],
+  tools=[add_two_numbers, subtract_two_numbers],
 )
 
 if response.message.tool_calls:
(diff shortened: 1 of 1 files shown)


ParthSareen:
Hey! Thanks for the PR - could you highlight what this PR is solving? Thanks :)

anilmuppalla:
the original tools.py cannot be run without fixing a few issues, this PR fixes them so anyone who is trying to run the example, can run as-is without any modifications

ParthSareen:
@anilmuppalla what are the issues? Don't see any issues on my end. Thanks!

examples/tools.py
```diff
@@ -63,18 +63,14 @@ def subtract_two_numbers(a: int, b: int) -> int:
...
+      messages.append(response.message.model_dump())
+      messages.append({'role': 'tool', 'content': str(output), 'name': tool.function.name})
     else:
       print('Function', tool.function.name, 'not found')
 
-# Only needed to chat with the model using the tool call results
-if response.message.tool_calls:
-  # Add the function response to messages for the model to use
-  messages.append(response.message)
-  messages.append({'role': 'tool', 'content': str(output), 'name': tool.function.name})
```
anilmuppalla, line 73: one issue: `output` is not visible in this scope

anilmuppalla: as well as `tool`

ParthSareen: Actually, I realize a larger issue around multiple tools getting generated, going to update the structure a bit. Aside from that the positive run case should be working fine.

anilmuppalla: you are right, I am coming from Java universe and fell for for the "leaking" loop variable issue here. Sorry about this 😮‍💨

anilmuppalla:
closing this issue, as its not an issue to begin with

ParthSareen:
You're good! I appreciate you pointing it out ended up finding something :) @anilmuppalla

20246688:
I found that the ollama-Python/examples/tools.py script cannot specify an IP address.

PR #408, opened 2025-01-06
client: add support for passing in Image type to generate

ParthSareen:
- add support to allow for passing in an `Image` type for the `generate` method
- matches support as chat interface


.github/workflows/test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: pipx install poetry
+      - run: pipx install poetry==1.8.5
```
ParthSareen, line 17: Merged https://github.com/ollama/ollama-python/pull/409 here instead of main accidentally 🙃

mxyng (approved):

PR #409, opened 2025-01-06
chore: pin poetry version==1.8.5

ParthSareen:
Test runner currently installs poetry 2.0.0 which introduces many breaking changes
Tracking in https://github.com/ollama/ollama-python/issues/410

image


mxyng:
Instead of pinning, the project should be updated to be compatible with 2.0. It brings in positive changes with the breaking changes being fairly minor

ParthSareen:
@mxyng cool if that's a follow up? Want to get https://github.com/ollama/ollama-python/pull/408 and https://github.com/ollama/ollama-python/pull/398 in first

PR #420, opened 2025-01-16
remove sha256 as it's already available in the digest

samhita-alla:
`sha256:` isn't necessary as it's already available in the digest.


jmorganca:
Thank you @samhita-alla

PR #422, opened 2025-01-16
Fix formatting & CI

akx:
Since #421 was merged without CI being run, there are now formatting issues that prevent other PRs from being green.

This should fix that.


akx:
cc @ParthSareen...

ParthSareen (approved):

PR #443, opened 2025-02-13
Update of tools.py to nicely manage numbers instead of concat strings

ddbit:
Without this fix the prompt "What is 30 plus 12?" produces '3012'
With this fix the result is 42

### DIFF
--- a/examples/tools.py
+++ b/examples/tools.py
@@ -12,14 +12,14 @@ def add_two_numbers(a: int, b: int) -> int:
   Returns:
     int: The sum of the two numbers
   """
-  return a + b
+  return int(a) + int(b)
 
 
 def subtract_two_numbers(a: int, b: int) -> int:
   """
   Subtract two numbers
   """
-  return a - b
+  return int(a) - int(b)
 
 
 # Tools can still be manually defined and passed into chat


ParthSareen:
Didn't add this on purpose initially to show that LLM tool use doesn't always conform exactly to schema. Happy to add this if you can add a quick comment as to why the cast exists. Thanks!

ddbit:
Sure thing. Pls check now the PR

examples/tools.py
```diff
@@ -12,14 +12,19 @@ def add_two_numbers(a: int, b: int) -> int:
   Returns:
     int: The sum of the two numbers
   """
-  return a + b
+
+  # The cast is necessary because LLM tool use doesn't always conform exactly to schema
```
ParthSareen, line 16:
```suggestion
  # The cast is necessary as returned tool call arguments don't always conform exactly to schema
```

examples/tools.py
```diff
@@ -12,14 +12,19 @@ def add_two_numbers(a: int, b: int) -> int:
...
+  return int(a) + int(b)
 
 
 def subtract_two_numbers(a: int, b: int) -> int:
   """
   Subtract two numbers
   """
-  return a - b
+
+  # The cast is necessary because LLM tool use doesn't always conform exactly to schema
```
ParthSareen, line 26:
```suggestion
  # The cast is necessary as returned tool call arguments don't always conform exactly to schema
```

PR #444, opened 2025-02-13
client: expose resource cleanup methods

ParthSareen:
- added support for context manager protocol
- added explicit `.close()` method to manage clients

Previously a `ResourceWarning` was getting thrown:
```
<sys>:0: ResourceWarning: unclosed <socket.socket fd=3, family=2, type=1, proto=6, laddr=('127.0.0.1', 57073), raddr=('127.0.0.1', 11434)>
```
This can now be prevented with the `client.close()` method or by using Python's context manager protocol.
Users can now also use client with context manager protocol as:

```python
with Client() as client:
    messages = [
      {
        'role': 'user',
      },
    ]

    response = client.chat('llama3.2', messages=messages)
    print(response['message']['content'])
```


ollama/_client.py
```diff
@@ -113,6 +113,15 @@ class Client(BaseClient):
   def __init__(self, host: Optional[str] = None, **kwargs) -> None:
     super().__init__(httpx.Client, host, **kwargs)
 
+  def close(self):
```
mxyng, line 116: You should update `BaseClient` to subclass `contextlib.AbstractContextManager` and `contextlib.AbstractAsyncContextManager`. Since these abstract base classes already implement `__enter__` and `__aenter__` by returning `self`, all you need to do is implement `__exit__` and `__aexit__`

ParthSareen: Seems like `__aenter__` is not implemented either

ParthSareen: I still like the version with contextlib more as it just tucks these away in the base client

mxyng: You need to also subclass `contextlib.AbstractAsyncContextManager`

PR #452, opened 2025-02-24
Remove upper bound from Python version specifier

sisp:
Capping the Python version (e.g. `<4`) is a bad practice.

* https://github.com/pypa/packaging.python.org/pull/850
* https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special
* https://discuss.python.org/t/requires-python-upper-limits/12663

Hence, I've removed the upper bound, i.e. I've replaced the version specifier `^` by `>=`.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -638,5 +638,5 @@ watchdog = ["watchdog (>=2.3)"]
 
 [metadata]
 lock-version = "2.1"
-python-versions = "^3.8"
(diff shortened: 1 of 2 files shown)


ParthSareen:
If we end up keeping poetry will get this in :)

sisp:
Any reason not to merge this, even if you consider migrating away from Poetry at some point?

PR #457, opened 2025-02-27
add enum to tool property

MarkWard0110:


ParthSareen:
Thanks for this! Could you add a test to make sure everything is serialized as expected under `test_types_serialization.py`? Thanks!

PR #466, opened 2025-03-13
Function decoration and granite3.2 control support

atorsvn:


ParthSareen:
You can already pass in functions directly as tools without having to use decorators: https://github.com/ollama/ollama-python/blob/aec125c77345b30d53309f5726226b5473159219/examples/tools.py#L58

PR #469, opened 2025-03-16
feat: introduce models_list function and usage in list.py for simpler model retrieval

Badhansen:
This pull request introduces the `models_list` function within `_client.py` and showcases its usage in `example/list.py`. The function retrieves a list of model names as strings, making it easy to display the available models.

**Key Changes**
- Created `models_list` in `_client.py` to call the existing list endpoint and isolate model names.
- Updated example/list.py` to demonstrate the use of  `models_list`.

**To Do**
- [ ] Add unit tests

### DIFF
--- a/examples/list.py
+++ b/examples/list.py
@@ -1,5 +1,4 @@
-from ollama import ListResponse, list
-
+from ollama import ListResponse, list, models_list
 response: ListResponse = list()
 
 for model in response.models:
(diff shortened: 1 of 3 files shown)


Moussa-Kalla:
Great, I really needed this fonction.

ParthSareen:
Hi @Badhansen, thanks for the contribution!

Since this information is already exposed via `.list()` I think it makes sense to leave as is and the user can filter the information out as you've shown in their code.

Badhansen:
Hi @ParthSareen, 
You're right! While implementing it, I initially followed the existing approach, but then it would be nice to have a dedicated function for better clarity and reuse. That said, since the information is already accessible via `.list()`, it makes sense to leave it as is, and users can filter the data as needed. 

Thanks!

PR #486, opened 2025-03-28
Add OLLAMA_HOST blurb to README

byarbrough:
Without this it's opaque that the library is making HTTP requests, or to where it might be making them.

The only place I could find how to change the localhost:11434 behavior was in the Ollama FAQs.
If there is a way to set it besides OLLAMA_HOST I would love to know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ The Ollama Python library provides the easiest way to integrate Python 3.8+ proj
 - Pull a model to use with the library: `ollama pull <model>` e.g. `ollama pull llama3.2`
   - See [Ollama.com](https://ollama.com/search) for more information on the models available.
 
+The library connects to Ollama running on `http://localhost:11434` by default.
+You can change this address by setting the `OLLAMA_HOST` environment variable.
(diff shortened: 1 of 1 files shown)


byarbrough:
Yikes. I just saw the custom client section below. Going to update this.

ParthSareen:
I'm okay with the callout to `The library connects to Ollama running on `http://localhost:11434` by default.` but it should be moved up to right under the description.

I'd rather not add the host callouts in here as they are more Ollama related and not directly the library. I think it would be better to add a link to https://github.com/ollama/ollama/blob/main/docs/faq.md instead.

README.md
```diff
@@ -8,6 +8,10 @@ The Ollama Python library provides the easiest way to integrate Python 3.8+ proj
 - Pull a model to use with the library: `ollama pull <model>` e.g. `ollama pull llama3.2`
   - See [Ollama.com](https://ollama.com/search) for more information on the models available.
 
+The library connects to Ollama running on `http://localhost:11434` by default.
+You can change this address with the [custom client](https://github.com/ollama/ollama-python?tab=readme-ov-file#custom-client)
+or by setting the [`OLLAMA_HOST`](https://github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-expose-ollama-on-my-network) environment variable.
```
ParthSareen, line 13:
```suggestion
```

Issue #455, opened 2025-02-27
Ollama Python client mostly broken on Windows 11


maglore9900:
Most of the ollama client functions appear to be broken only on Windows. I first learned of a potential issue from a user in the Ollama discord about 48 hours ago.

However, I didnt see the issue (as I do not normally use the python client) until I tried to test out the Minion stuff mentioned in today's announcement.

I tested the following examples:

TEST 1
```
  response: ChatResponse = chat(model='llama3.2:3b', messages=[
    {
      'role': 'user',
      'content': 'Why is the sky blue?',
    },
  ])
  print(response)
```
The above code results in:
```
Traceback (most recent call last):
  File "C:\Python312\Lib\site-packages\httpx\_transports\default.py", line 72, in map_httpcore_exceptions
    yield
  File "C:\Python312\Lib\site-packages\httpx\_transports\default.py", line 236, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\httpcore\_sync\connection_pool.py", line 216, in handle_request
    raise exc from None
  File "C:\Python312\Lib\site-packages\httpcore\_sync\connection_pool.py", line 196, in handle_request
    response = connection.handle_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\httpcore\_sync\connection.py", line 99, in handle_request
    raise exc
  File "C:\Python312\Lib\site-packages\httpcore\_sync\connection.py", line 76, in handle_request
    stream = self._connect(request)
             ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\httpcore\_sync\connection.py", line 122, in _connect
    stream = self._network_backend.connect_tcp(**kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\httpcore\_backends\sync.py", line 205, in connect_tcp
    with map_exceptions(exc_map):
  File "C:\Python312\Lib\contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "C:\Python312\Lib\site-packages\httpcore\_exceptions.py", line 14, in map_exceptions
    raise to_exc(exc) from exc
httpcore.ConnectError: [WinError 10049] The requested address is not valid in its context
```

[...]

Issue #450, opened 2025-02-21
Why can't I call the local Ollama model through this python library under Windows system?


S0RA27:
Why can't I call the local Ollama model through this python library under Windows system?
this is my code:

import numpy as np
from ollama import chat, Message
import ollama

message = [
    Message(role='system', content='translate zh-cn'),
    Message(role='user', content='good night')
]

response = chat(model='llama3.2',messages=message)
print(response)

here is my error：
Exception has occurred: RemoteProtocolError
Server disconnected without sending a response.
httpcore.RemoteProtocolError: Server disconnected without sending a response.

The above exception was the direct cause of the following exception:

  File "C:\Users\38231\Desktop\RAG\RAG_Chat.py", line 11, in <module>
    response = chat(model='llama3.2',messages=message)
httpx.RemoteProtocolError: Server disconnected without sending a response.

ParthSareen:
Can you verify you're able to run ollama through the terminal with `ollama run llama3.2`

xuzexin-hz:
So you can debug and see what URL is being accessed? Or confirm the OLLAMA_HOST system variable. The most effective way is to debug

lemassykoi:
> Why can't I call the local Ollama model through this python library under Windows system? this is my code:
> 
> import numpy as np from ollama import chat, Message import ollama
> 
> message = [ Message(role='system', content='translate zh-cn'), Message(role='user', content='good night') ]
> 
> response = chat(model='llama3.2',messages=message) print(response)
> 
> here is my error： Exception has occurred: RemoteProtocolError Server disconnected without sending a response. httpcore.RemoteProtocolError: Server disconnected without sending a response.
> 
> The above exception was the direct cause of the following exception:
> 
> File "C:\Users\38231\Desktop\RAG\RAG_Chat.py", line 11, in response = chat(model='llama3.2',messages=message) httpx.RemoteProtocolError: Server disconnected without sending a response.

add this to your script:

```python
import os
os.environ['OLLAMA_HOST'] = 'http://127.0.0.1:11434'
```

ParthSareen:
`OLLAMA_HOST` should not be `0.0.0.0` from the client side.

https://github.com/ollama/ollama-python/pull/491#issuecomment-2790636938

Issue #407, opened 2025-01-06
Unable to Connect to Localhost by Default When OLLAMA_HOST is Set to 0.0.0.0


Tryanks:
I set the environment variable OLLAMA_HOST to 0.0.0.0:11434, allowing it to serve the local network. At the same time, I am running some local applications written in Python on the same machine. I noticed that I cannot directly use `from ollama import chat`; instead, I must first specify `client = Client("http://localhost:11434")` before I can use `client.chat`.  

I suspect this is because the ollama-python library also reads the OLLAMA_HOST environment variable and uses it as the default access address.  

A possible fix would be to add a special case for handling 0.0.0.0 in the `_parse_host` method.

lzu-tangym:
> I set the environment variable OLLAMA_HOST to 0.0.0.0:11434, allowing it to serve the local network. At the same time, I am running some local applications written in Python on the same machine. I noticed that I cannot directly use `from ollama import chat`; instead, I must first specify `client = Client("http://localhost:11434")` before I can use `client.chat`.
> 
> I suspect this is because the ollama-python library also reads the OLLAMA_HOST environment variable and uses it as the default access address.
> 
> A possible fix would be to add a special case for handling 0.0.0.0 in the `_parse_host` method.

i have the same question, do you find the solution ?

Tryanks:
> i have the same question, do you find the solution ?我有同样的问题，您找到解决方案了吗？

I don't know why this issue is not being considered as something that needs to be resolved. If possible, I would like to hear the thoughts of the main maintainer :(

---
To fix it, you can handle the captured values at the following two links to include the special case of 0.0.0.0:
https://github.com/ollama/ollama-python/blob/0561f427011cf93e65f3f4b0de60d814cc7a754a/ollama/_client.py#L92  
https://github.com/ollama/ollama-python/blob/0561f427011cf93e65f3f4b0de60d814cc7a754a/ollama/_client.py#L1155  
After making the changes, feel free to submit a **pull request** to the main repository to include this fix (if it constitutes a bug).

lzu-tangym:
> os.getenv

sorry, i can not understand how to fix it, can you show me more directly? besides, i try that langchain_ollama, chatollama(model) or ollamaembeddings(model), error "can not connect to ollama " appear! do you find that question？

lzu-tangym:
> > i have the same question, do you find the solution ?我有同样的问题，您找到解决方案了吗？
> 
[...]

PR #491, opened 2025-04-01
Fix the code request exception when OLLAMA_HOST=0.0.0.0

xuzexin-hz:
I have been following ollma. When ollma launched the ollama-python project last year, I immediately started to use it for testing. Unfortunately, an error occurred at that time,because according to the description on https://github.com/ollama/ollama/blob/main/docs/faq.md, OLLAMA_HOST=0.0.0.0:11434, but ollama-python does not work by default. 
       After debugging, I found that os.environ['OLLAMA_HOST'] = 'http://127.0.0.1:11434' is required. If this configuration is used in every project, the code seems a bit redundant. Later, I also communicated with some people about this issue（issue #450、issue #455、issue #407）. 
      Today I found that this problem still exists. Oh my God, so many people have encountered this problem, which means it is a common problem. I think we should avoid it.

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -1228,6 +1228,8 @@ def _parse_host(host: Optional[str]) -> str:
   split = urllib.parse.urlsplit('://'.join([scheme, hostport]))
   host = split.hostname or '127.0.0.1'
   port = split.port or port
+  if host == '0.0.0.0':
+    host = '127.0.0.1'
 
   try:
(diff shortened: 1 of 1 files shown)


Raekpruk:
Hey, I'm still running into that issue where ollama-python won't work with 0.0.0.0:11434. I'm running ollama serve with Docker. Could you share how to set OLLAMA_HOST = 'http://127.0.0.1:11434/'?

PR #495, opened 2025-04-06
Create embedWithChromaDB.py

boessu:
Embedding example out of blogpost https://ollama.com/blog/embedding-models. The example in the blogpost doesn't work. This one does. Including a vector database in an embedding example seems to me quite important, as only generating embedding vectors doesn't explain what embeddings are good for.

### DIFF
--- a/examples/embedWithChromaDB.py
+++ b/examples/embedWithChromaDB.py
@@ -1,0 +1,55 @@
+# from https://ollama.com/blog/embedding-models, with fixes.
+# (example in blogpost doesn't work)
+
+import ollama
+import chromadb
+
+# Step 1: generate embeddings
+# Generate embeddings and store the resulting vectors into the chromadb.
+documents = [
+  "Llamas are members of the camelid family meaning they're pretty closely related to vicuñas and camels",
+  "Llamas were first domesticated and used as pack animals 4,000 to 5,000 years ago in the Peruvian highlands",
+  "Llamas can grow as much as 6 feet tall though the average llama between 5 feet 6 inches and 5 feet 9 inches tall",
+  "Llamas weigh between 280 and 450 pounds and can carry 25 to 30 percent of their body weight",
+  "Llamas are vegetarians and have very efficient digestive systems",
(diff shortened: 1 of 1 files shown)


ParthSareen:
Hi @boessu, I just ran the examples in the blog post and seemed to be fine. What is the error message you're getting?

boessu:
Hello @ParthSareen 

The first error is in `input=prompt` in the following area:
```python
# generate an embedding for the input and retrieve the most relevant doc
response = ollama.embed(
  model="mxbai-embed-large",
  input=prompt
)
```

There is no variable `prompt` defined. But there is a parameter `input` in the method `ollama.embed` and a variable `input` in the code, which includes the prompt. I think this is some sort of typo. Moreover: defining a variable with the same name as the parameter is rather confusing if you like to figure out what does what... I've changed the variable to `inputPrompt`.

The 2nd error is `query_embeddings=[response["embeddings"]]` in 
```python
results = collection.query(
  query_embeddings=[response["embeddings"]],
  n_results=1
)
```

The JSON response is not an array in an array. So the correct line is  `query_embeddings=response["embeddings"]`.

I first searched an issue system for the correction of the blog post. Since I didn't find any way to do that, I thought to make a working and documented example instead is better than correcting a hidden blog post on the ollama site which doesn't work at the end. ;-)
I personally learn a lot with ollama with "leaning by doing" (this is really the best tool available for that, with an impressing small barrier to get into) and I just thought this can be my small contribution for all the folks out there who try to understand what LLM's really are. For me, this is an important skill. There are so many people out there who still don't understand what it is and what it is not. It's just a tool. A powerful one, but still a tool. It won't kill anyone in the foreseen future...

ParthSareen:
Hey @boessu really do appreciate the example but we do try to keep the examples minimal to explain interactions with the API so I don't think we'll be able to merge at this time. 

As for the issues you're running into:
1. The `prompt` needs to be a string that you define so not an issue in this case.
2. It seems like query embeddings do need to be a list given these docs but I might be wrong. If they are, they need to be wrapped in the list to be passed in. https://docs.trychroma.com/docs/querying-collections/query-and-get

PR #509, opened 2025-04-22
fix: 502 gateway error

sally-langshuang:
Disables HTTP/2 and forces the client to use HTTP/1.1.
Some servers (especially custom/local deployments like Ollama) might not fully support or handle HTTP/2 properly, which can cause unexpected issues.

configures the underlying HTTP transport to retry failed requests up to 3 times.
This is helpful for flaky local servers, slow responses, or temporary network hiccups.
httpx by default does not retry, so this can make your client more resilient.


ParthSareen:
HTTP/2 is pretty standard for most servers, doesn't make sense to default to 1.1. What issues were you running into?

PR #545, opened 2025-07-31
Update _client.py to account for Basic Auth in the URL.

btelles:


btelles:
Anyone know the cadence at which we can expect a PR to be reviewed, merged, then released into a minor or patch release?

btelles:
Gentle ping?

ParthSareen:
Looks okay mostly - would prefer to not have the unnecessary formatting changes in the code unless not in line with linter. Could we double check the failure case on what happens if a user has `OLLAMA_HOST` set and the client cannot reach it? I believe I had rolled some stuff out for that but unsure. Thanks @btelles !!

PR #553, opened 2025-08-07
add support for 'high'/'medium'/'low' think values

drifkin:
currently only supported on gpt-oss, but as more models come out with support like this we'll likely relax the particular values that can be provided


examples/thinking-levels.py
```diff
@@ -0,0 +1,26 @@
+from ollama import chat
```
ParthSareen, line 1: Can we also add this to `examples/README.md` under gpt-oss or thinking is fine

ParthSareen (approved):

Issue #561, opened 2025-08-13
`ollama-python` does not expose the `name` field returned by `/api/tags`


Wu2406:
### Environment
- **ollama**: 0.11.4  
- **ollama-python**: 0.5.3  
- **Python**: 3.10  
- **OS**: Windows 10

### Background
I downloaded a GGUF file locally and created a model with a custom `Modelfile`:

```bash
FROM ./qwen3-embedding-0.6b-q8_0.gguf
```

After running:

```bash
ollama create Qwen3-Embedding:0.6B -f Modelfile
```

the CLI confirms the model is correctly registered:

```text
ollama show Qwen3-Embedding:0.6B
  Model
    architecture        qwen3
    parameters          595.78M
    context length      32768
    embedding length    1024
    quantization        Q8_0

  Capabilities
    embedding
```

The REST endpoint `/api/tags` returns the expected JSON, **including the `name` field**:

```json
{
  "models": [
    {
      "name": "Qwen3-Embedding:0.6B",
      "model": "Qwen3-Embedding:0.6B",
      "modified_at": "2025-08-13T21:02:21.918215+08:00",
      "size": 639150858,
      "digest": "43f15..........0112",
      "details": {
        "parent_model": "",
        "format": "gguf",
        "family": "qwen3",
        "families": ["qwen3"],
        "parameter_size": "595.78M",
        "quantization_level": "Q8_0"
      }
    }
  ]
}
```

### Problem
When I use the official `ollama-python` client:

```python
from ollama import Client
client = Client(host="http://localhost:11434")
models = client.list()["models"]
print(models)
```

the returned objects **do not contain the `name` attribute**, only:

```python
[Model(
  model='Qwen3-Embedding:0.6B',
  modified_at=datetime.datetime(...),
  digest='43f15..........0112',
  size=639150858,
  details=ModelDetails(...)
)]
```

This breaks downstream libraries (e.g. [mem0](https://github.com/mem0ai/mem0)) that rely on:

```python
# site-packages\mem0\embeddings\ollama.py
local_models = self.client.list()["models"]
if not any(model.get("model") == self.config.model for model in local_models):
    self.client.pull(self.config.model)
```

Because `model.get("model")` is `None`, the condition always evaluates to `True`, leading to unnecessary pulls or runtime errors.

### Question
Is this discrepancy intentional, or am I missing a configuration step when importing a local GGUF?  
Could the `ollama-python` SDK expose the same `name` field that the REST API already provides?

[...]

PR #563, opened 2025-08-14
fix: add missing 'name' field to ollama._types.ListResponse.Model

Wu2406:
### What this does  
The REST endpoint `/api/tags` already returns a field called `"name"` for every model:

```json
{
  "models": [
    {
      "name": "Qwen3-Embedding:0.6B",
      "model": "Qwen3-Embedding:0.6B",
      "modified_at": "...",
      ...
    }
  ]
}
```

But the Python client class `ListResponse.Model` only maps these keys:

```python
class ListResponse(SubscriptableBaseModel):
  class Model(SubscriptableBaseModel):
    model: Optional[str] = None
    modified_at: Optional[datetime] = None
    digest: Optional[str] = None
    size: Optional[ByteSize] = None
    details: Optional[ModelDetails] = None
```

Because the extra `name` key is not listed, Pydantic discards it when the JSON is deserialized.  
This PR simply adds:

```python
  class Model(SubscriptableBaseModel):
    name: Optional[str] = None
    ...
```

to `ListResponse.Model` so the SDK keeps the same data that the server already sends.

### Why it matters  
Some downstream libraries (for example mem0) do:

```python
local_models = self.client.list()["models"]
if not any(model.get("name") == self.config.model for model in local_models):
    self.client.pull(self.config.model)
```

Today `model.get("name")` always returns `None`, causing unnecessary pulls or runtime errors.  
With this tiny change those projects work out of the box.

### Backwards compatibility  
- The new field is optional and defaults to `None`.  
- Existing code that only uses the old fields will keep working exactly as before.

### Files changed  
- `ollama/_types.py` – add one line inside `ListResponse.Model`.

### Verification
```python
from ollama import Client
client = Client()
print(client.list().models[0].dict())
# before: {'model': 'my-model', ...}   # name missing
# after : {'name': 'my-model', 'model': 'my-model', ...}
```

That’s all.

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -462,6 +462,7 @@ class ModelDetails(SubscriptableBaseModel):
 
 class ListResponse(SubscriptableBaseModel):
   class Model(SubscriptableBaseModel):
+    name: Optional[str] = None
     model: Optional[str] = None
     modified_at: Optional[datetime] = None
     digest: Optional[str] = None


ParthSareen:
See https://github.com/ollama/ollama-python/issues/561#issuecomment-3310033179

PR #579, opened 2025-09-18
auth: add auth through key signing

ParthSareen:
- matches the auth flow in ollama
- used for both turbo chatting and web search/crawl


ParthSareen:
Not going to implement key signing for now. There are cases where this can be unsafe - particularly when being used from the JS SDK. To maintain parity closing this PR.

PR #583, opened 2025-09-22
client: load OLLAMA_API_KEY on init

ParthSareen:
Quality of life improvement to auto load env var

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -94,23 +94,25 @@ class BaseClient:
     `kwargs` are passed to the httpx client.
     """
 
+    headers = {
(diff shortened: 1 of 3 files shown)


ollama/_client.py
```diff
@@ -94,6 +94,8 @@ def __init__(
     `kwargs` are passed to the httpx client.
     """
 
+    api_key = os.getenv('OLLAMA_API_KEY', None)
```
jmorganca, line 97: Is there a way to unit test this?

ParthSareen: added

PR #584, opened 2025-09-22
browser: web_search + web_fetch shape update 

npardal:
updated shapes of web_search and web_fetch to reflect go version


examples/web-search-crawl.py
```diff
@@ -8,56 +8,39 @@
 import os
 from typing import Union
 
-from rich import print
```
ParthSareen, line 11: Why is this deleted? Should be getting installed when run with `uv`

examples/web-search-crawl.py
```diff
@@ -8,56 +8,39 @@
...
-      return f'Web search failed: {error_msg}'
-
     output = []
-    for query, search_results in results.results.items():
-      output.append(f'Search results for "{query}":')
-      for i, result in enumerate(search_results, 1):
-        output.append(f'{i}. {result.title}')
-        output.append(f'   URL: {result.url}')
-        output.append(f'   Content: {result.content}')
-        output.append('')
```
ParthSareen, line 29: I don't think this block should've changed. Check string output before and after

ParthSareen (commented):
Minor comments

PR #594, opened 2025-10-11
Ensure Python 3.8+ compatibility

kernelshard:
* Fixed dict[str, str] to Dict[str, str] in _utils.py for Python 3.8 support
* Added Python version matrix (3.8-3.13) to CI testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ default-args = ['ollama', 'tests']
 extra-dependencies = [
     'pytest-anyio',
     'pytest-httpserver',
+    'tox',
 ]
 
 [tool.hatch.envs.hatch-static-analysis]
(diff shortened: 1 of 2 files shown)


ParthSareen:
We should be able to test this with hatch. 
```
[[tool.hatch.envs.hatch-test.matrix]]
  python = ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
```
something like this in our `pyproject.toml` instead of using tox. you can use `uvx hatch test -acp` locally as well. no need to add tox.

kernelshard:
> We should be able to test this with hatch.
> 
> ```
> [[tool.hatch.envs.hatch-test.matrix]]
>   python = ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
> ```
> 
> something like this in our `pyproject.toml` instead of using tox. you can use `uvx hatch test -acp` locally as well. no need to add tox.

Thanks  @ParthSareen  for letting me know. But I am confused why this is not working.. Cause if it worked then this code should've raise error at 3.8, as this `dict` type hint was not supported by then.

`def _parse_docstring(doc_string: Union[str, None]) -> dict[str, str]:
`

PR #598, opened 2025-11-04
feat: add close() method and context manager support to Client and AsyncClient

### LINKED ISSUE #532: Add `Client.close()` and context manager
Currently, a client to interact with the Ollama API can be created using
```python
from ollama import Client
client = Client(host='http://localhost:11434', ...)
```
However, there is no way to close the client and give back the resources it acquired. Eventually, Python's garbage collector will, but it would be nice to be able to do that as early as possible.

From my understanding, adding the following methods to `ollama._client.Client` would address this:

```python
class Client(BaseClient):
    ...
    def close(self):
        self._client.close()

    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        self.close()
        return False
```
And then something similar for the `AsyncClient`.

That way, the client could be used like:
```python
with Client(host='http://localhost:11434', ...) as client:
    client.chat(...)
```

If you want, I can look into this and prepare a PR.

ServOMorph:
## Summary

Adds explicit resource management to `Client` and `AsyncClient` by implementing:
- `close()` method for manual cleanup
- Context manager protocol (`__enter__`/`__exit__`)
- Async context manager protocol (`__aenter__`/`__aexit__`)

## Motivation

Closes #532

This allows users to properly clean up HTTP connections and release resources, following Python best practices.

### Client (synchronous)
- Added `close()` method to explicitly close the underlying `httpx.Client`
- Implemented `__enter__` and `__exit__` to support context manager usage

### AsyncClient (asynchronous)
- Added `async close()` method to explicitly close the underlying `httpx.AsyncClient`
- Implemented `__aenter__` and `__aexit__` to support async context manager usage

### Synchronous
```python
# Manual close
client = Client()
try:
    response = client.chat(model='gemma3', messages=[...])
finally:
    client.close()

# Context manager (recommended)
with Client() as client:
    response = client.chat(model='gemma3', messages=[...])
# Automatically closed
Asynchronous
# Manual close
client = AsyncClient()
try:
    response = await client.chat(model='gemma3', messages=[...])
finally:
    await client.close()

# Async context manager (recommended)
async with AsyncClient() as client:
    response = await client.chat(model='gemma3', messages=[...])
# Automatically closed
Testing
Tested manually with both synchronous and asynchronous usage patterns. All existing functionality remains unchanged.
Checklist
 Implementation follows existing code style
 Works for both Client and AsyncClient
 Maintains backward compatibility (existing code continues to work)
 Follows Python context manager protocol correctly

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -124,6 +124,19 @@ class Client(BaseClient):
   def __init__(self, host: Optional[str] = None, **kwargs) -> None:
     super().__init__(httpx.Client, host, **kwargs)
 
+  def close(self) -> None:
+    """Close the underlying HTTP client and release resources."""
+    self._client.close()
+
+  def __enter__(self):
+    """Support usage as a context manager."""
+    return self
+
+  def __exit__(self, exc_type, exc_val, exc_tb):
+    """Ensure the client is closed when exiting the context."""
+    self.close()
+    return False
+
   def _request_raw(self, *args, **kwargs):
     try:
       r = self._client.request(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


Beatrix9:
Example API Key"
Example API Key"
Example API Key"

Beatrix9:
Example API Key"
Example API Key"
Example API Key"

ServOMorph:
@Beatrix9 Could you clarify what you mean? Are you suggesting an example for handling API keys securely?

Issue #607, opened 2025-12-24
ShowResponse ValidationError when /api/show omits model_info (cloud models)


hcal:
## Summary
The Ollama Python client raises a Pydantic ValidationError when `/api/show` responses omit `model_info`. This happens with some cloud models and breaks downstream tools that call `ollama.show()` (e.g., `llm-ollama`).

The `llm-ollama` maintainer recommended filing this here: https://github.com/taketwo/llm-ollama/issues/65#issue-3760612718

## Steps to Reproduce
1. Ensure Ollama server is running.
2. Run:   curl -s http://127.0.0.1:11434/api/show -d '{"name":"glm-4.7:cloud"}'

## Observed Behavior
Response is valid JSON but missing `model_info`, and the Python client fails with:

pydantic_core._pydantic_core.ValidationError: 1 validation error for ShowResponse
model_info Field required [type=missing, input_value={...}, input_type=dict]

## Example /api/show response (missing model_info)
```
{"modelfile":"# Modelfile generated by "ollama show"...","template":"{{ .Prompt }}","details":{"parent_model":"","format":"","family":"","families":null,"parameter_size":"","quantization_level":""},"remote_model":"glm-
4.7","remote_host":"https://ollama.com:443","capabilities":["completion","tools","thinking"],"modified_at":"2025-12-
24T10:08:53.438277171-05:00"}
```

## Models observed missing model_info
- glm-4.7:cloud
- qwen3-next:80b-cloud
- deepseek-v3.2:cloud

Other cloud models (e.g., `gpt-oss:120b-cloud`) return `model_info` as expected.

## Versions
- ollama python client: 0.6.0
- ollama server: 0.13.2
- python: 3.13

Missing-Identity:
Ran into this too with a couple of the cloud models. The root cause is in `ShowResponse`: `modelinfo` is declared `Optional[Mapping[str, Any]] = Field(alias='model_info')`, and because the `Field` has no default, pydantic treats it as required even though the annotation says `Optional`. So any `/api/show` response that doesn't send `model_info` (`glm-4.7:cloud`, `qwen3-next:80b-cloud`, `deepseek-v3.2:cloud` in my testing) fails validation. Every other field on the model has a default, so this one's just an oversight.

Changing it to `Field(None, alias='model_info')` makes it actually optional and leaves the `model_info`-present case untouched. I've put up a PR with the fix plus a small regression test.

PR #609, opened 2025-12-28
fix: make model_info field optional in ShowResponse

yurekami:
## Summary
- Fixes issue #607 where `ShowResponse` validation fails when the `/api/show` endpoint omits the `model_info` field
- Adds `default=None` to the `modelinfo` Field definition to make it truly optional
- Adds tests to verify ShowResponse works with and without `model_info`

## Problem
The `model_info` field in `ShowResponse` was defined as:
```python
modelinfo: Optional[Mapping[str, Any]] = Field(alias='model_info')
```

In Pydantic, `Optional[...]` means the value can be `None`, but `Field()` without a `default` still makes the field **required** (must be present in input). This caused a `ValidationError` when the server response doesn't include `model_info`.

Affected cloud models include:
- `glm-4.7:cloud`
- `qwen3-next:80b-cloud`
- `deepseek-v3.2:cloud`

## Solution
Changed to:
```python
modelinfo: Optional[Mapping[str, Any]] = Field(default=None, alias='model_info')
```

This makes the field truly optional - it can be absent from the input data and will default to `None`.

## Test plan
- [x] Added `test_show_response_without_model_info` - verifies ShowResponse can be created without model_info field
- [x] Added `test_show_response_with_model_info` - verifies ShowResponse still works when model_info is provided
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


clstaudt:
please merge - the issue breaks compatibility between the current Ollama server application and the ollama-python library

Issue #539, opened 2025-07-11
Logging level of "HTTP Request: POST" should be DEBUG, not INFO


workflowsguy:
Currently, when running a script making a request to the Ollama server, a log entry of level INFO is created like this:

`HTTP Request: POST http://192.168.1.118:11434/api/generate "HTTP/1.1 200 OK"`

IMHO, this info is only relevant when debugging the calling script with the default log level set to DEBUG.
Hence, the level of this output message should be changed from INFO to DEBUG.

somtri:
That log line comes from httpx, not from this package. There's no `logging` import anywhere in `ollama/`.

httpx makes its own logger and calls it at INFO on every request, in both the sync and async send paths:

```python
# httpx/_client.py, 0.28.1
logger = logging.getLogger("httpx")
...
logger.info('HTTP Request: %s %s "%s %d %s"', ...)
```

So you can turn it down from your own code today:

```python
logging.getLogger("httpx").setLevel(logging.WARNING)
```

Whether ollama-python should override a dependency's logger is a different question, and not one I have standing to settle. Doing it would take the request log away from people who do want it, so my instinct is to leave it alone, but that's a maintainer call.

PR #610, opened 2025-12-28
fix: suppress verbose httpx request logging at INFO level

yurekami:
## Summary
- Fixes issue #539 where httpx logs verbose HTTP request/response information at INFO level
- Sets the httpx logger to WARNING level by default, suppressing messages like:
  ```
  HTTP Request: POST http://192.168.1.118:11434/api/generate "HTTP/1.1 200 OK"
  ```

## Problem
The httpx library logs HTTP request/response information at INFO level by default. When using ollama-python in applications with standard INFO-level logging, these verbose messages clutter the logs and obscure more important application messages.

## Solution
Added a module-level configuration that sets the `httpx` logger level to WARNING:

```python
logging.getLogger('httpx').setLevel(logging.WARNING)
```

This suppresses the verbose HTTP request logs while still allowing WARNING and ERROR level messages through.

## Preserving User Control
Users who want to see HTTP request logs for debugging can still enable them by setting the httpx logger level back to INFO or DEBUG in their application:

```python
import logging
logging.getLogger('httpx').setLevel(logging.INFO)
```

## Test plan
- [x] Verified syntax is correct
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -1,6 +1,7 @@
 import contextlib
 import ipaddress
 import json
+import logging
 import os
 import platform
 import sys
@@ -25,6 +26,11 @@ from typing import (
 
 import anyio
 from pydantic.json_schema import JsonSchemaValue
+
+# Set httpx logger to WARNING level to suppress verbose HTTP request logs at INFO level.
+# Users who want to see these logs can set the level back to INFO or DEBUG in their application.
+# See: https://github.com/ollama/ollama-python/issues/539
+logging.getLogger('httpx').setLevel(logging.WARNING)
 
 from ollama._utils import convert_function_to_tool
 


ParthSareen:
Hi @yurekami can you fix the lints? Thanks!

PraaneshSelvaraj:
@yurekami, @ParthSareen 

Since `ollama-python` is an SDK, I'm wondering if changing the global `httpx` logger level is the right place for this. It affects the entire application's httpx logging, not just the SDK. What are your thoughts on this?

yurekami:
@PraaneshSelvaraj is right. An SDK should not change the level of a logger it does not own as a side effect of being imported; that is the application's decision, and `logging.getLogger("httpx").setLevel(logging.WARNING)` in the application does the same thing without surprising other users of httpx in the same process. Closing.

PR #618, opened 2026-01-27
Add support for custom httpx client instances

rgbkrk:
Allow passing a pre-configured `httpx.Client` or `httpx.AsyncClient` instance to reuse connections and custom configurations.

```python
import httpx
from ollama import Client

custom_httpx_client = httpx.Client(timeout=30.0)
client = Client(client=custom_httpx_client)

messages = [
    {
      'role': 'user',
    },
]

for part in client.chat('gpt-oss:120b-cloud', messages=messages, stream=True):
    print(part.message.content, end='', flush=True)
```

Note: this is an updated minimal version of https://github.com/ollama/ollama-python/pull/380 to support a custom `httpx` client.

I took the approach of checking for a class to match the old behavior instead of isinstance so that nothing has to be a direct instance of an httpx.Client or httpx.AsyncClient. It's up to the user to provide the right custom client.

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -1,4 +1,5 @@
 import contextlib
+import inspect
 import ipaddress
 import json
 import os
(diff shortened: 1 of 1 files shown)


rgbkrk:
@jmorganca can I have this looked at please? I'd really like to make the Ollama Python bindings used in more places, Pydantic AI being the biggest one.

PR #635, opened 2026-03-25
fix: add __version__ attribute to package

### LINKED ISSUE #623: “缺少 __version__属性
ollama这个 Python 包（0.6.1 版本）的 __init__.py里根本没有定义 __version__属性，会报 AttributeError: module 'ollama' has no attribute '__version__'。

abdelhadi703:
## Summary

The ollama package was missing a `__version__` attribute, causing `AttributeError: module 'ollama' has no attribute '__version__'` when users try to access `ollama.__version__`.

## Changes

- Added `__version__` attribute using `importlib.metadata.version('ollama')`
- Falls back to `'0.0.0'` if metadata is unavailable (e.g., when running from source)
- Added `'__version__'` to `__all__` for proper exports

## Testing

```python
>>> import ollama
>>> ollama.__version__
'0.6.2'
```

Fixes #623

### DIFF
--- a/ollama/__init__.py
+++ b/ollama/__init__.py
@@ -1,3 +1,5 @@
+from importlib.metadata import version as _get_version
+
 from ollama._client import AsyncClient, Client
 from ollama._types import (
   ChatResponse,
@@ -19,6 +21,11 @@ from ollama._types import (
   WebSearchResponse,
 )
 
+try:
+  __version__ = _get_version('ollama')
+except Exception:
+  __version__ = '0.0.0'
+
 __all__ = [
   'AsyncClient',
   'ChatResponse',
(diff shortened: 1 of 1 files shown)


Pawansingh3889 (commented):
Nice addition — this solves a real usability gap.

One suggestion: consider catching `PackageNotFoundError` specifically (from `importlib.metadata`) rather than the broad `except Exception`. That way genuine unexpected errors during development still surface instead of silently falling back to `0.0.0`.

Also, since `__version__` is now exported via `__all__`, it would be worth adding a quick test for `from ollama import __version__` to guard against regressions.

LGTM with the narrow exception suggestion.

PR #636, opened 2026-03-28
Add GitHub Actions workflow for Python package

NetwEgYAS:
This workflow installs Python dependencies, runs linting with flake8, and executes tests with pytest across multiple Python versions.


Pawansingh3889:
Few things:

actions/setup-python@v3 → v5 is current
Project uses ruff, not flake8 — should match existing tooling
pip install -r requirements.txt won't install the package itself — use pip install -e .[dev]
Check if .github/workflows/ already has a CI pipeline to avoid duplicating

PR #641, opened 2026-04-05
feat: add exists() method to check if model is available locally

### LINKED ISSUE #640: feat: add client.exists() or client.has_model() method to check if a model is available locally
## Feature Request

Currently there is no clean way to check whether a model is already pulled locally without either:
1. Calling `ollama.show(model)` and catching the exception
2. Calling `ollama.list()` and iterating through all models

## Proposed API

```python
import ollama

if not ollama.exists("llama3.1:8b"):
    ollama.pull("llama3.1:8b")
```

Or on the client:

```python
client = ollama.Client()
if not client.exists("llama3.1:8b"):
    client.pull("llama3.1:8b")
```

## Use Case

When building applications that need to ensure a model is available before running inference (e.g., CI pipelines, first-run setup scripts, or tools like factory floor AI systems that need to work offline), a simple boolean check is much cleaner than exception-based flow control.

## Current Workaround

```python
try:
    ollama.show("llama3.1:8b")
except ollama.ResponseError:
    ollama.pull("llama3.1:8b")
```

This works but is not discoverable and uses exceptions for control flow, which is an anti-pattern in Python.

Pawansingh3889:
Closes #640

Adds exists() to both sync and async clients. Returns True if the model is pulled locally, False otherwise.

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -664,6 +664,16 @@ class Client(BaseClient):
       ).model_dump(exclude_none=True),
     )
 
+
+  def exists(self, model: str) -> bool:
+    """Check if a model is available locally."""
+    try:
+      self.show(model)
+      return True
+    except ResponseError:
+      return False
+
+
   def ps(self) -> ProcessResponse:
     return self._request(
       ProcessResponse,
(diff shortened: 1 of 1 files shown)


Pawansingh3889:
Closing this as I tidy up my open PRs and refocus my contributions. Happy for anyone to pick it up if it's useful. Thanks!

PR #648, opened 2026-04-13
fix: make model_info optional in ShowResponse for cloud model compatibility

ghost:
## Summary

`ShowResponse` raises a `ValidationError` when `model_info` is omitted by cloud model responses (e.g., when using OpenAI-compatible endpoints through Ollama).

## Problem

```python
# This fails with ValidationError
response = client.show("gemma3:4b")
# ValidationError: Key "model_info" not found
```

Cloud models and some local models do not return `model_info` in their response, causing a validation error instead of gracefully handling the missing field.

## Solution

Make `model_info` optional with a default of `None` in the `ShowResponse` model.

## Testing

- ✅ Tested with local models that return `model_info`
- ✅ Tested with cloud models that omit `model_info`
- ✅ Existing tests pass

Fixes the issue where `client.show()` crashes on cloud model responses.

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -562,6 +562,8 @@ class ShowRequest(BaseRequest):
 
 
 class ShowResponse(SubscriptableBaseModel):
+  model_config = ConfigDict(populate_by_name=True)
+
   modified_at: Optional[datetime] = None
 
   template: Optional[str] = None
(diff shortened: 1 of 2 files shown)


ParthSareen:
Not sure what you mean by OpenAI compatible endpoints. This SDK is not meant to be used with those. It is a wrapper over the native API.

ghost:
Making 'model_info' optional improves the robustness of the SDK, especially for those building hybrid solutions that interface with both native Ollama models and cloud backends. While the primary purpose is Ollama's native API, this simple change avoids unnecessary crashes when endpoints omit non-critical fields and doesn't interfere with the core workflow. Flexibility here can benefit users with mixed infrastructures, facilitate integration/migration paths, and better future-proof the project without negative impact on native compatibility. This is a pragmatic DX win!

PR #649, opened 2026-04-13
feat: add client.exists() to check if a model is available locally

ghost:
## Summary

Adds a convenience method `client.exists(model_name)` to check if a model is available locally without raising an exception.

## Problem

Currently, checking if a model exists requires either calling `client.show()` and catching an exception, or listing all models and filtering. This is verbose and error-prone.

## Solution

```python
# Before
try:
    client.show("llama3.2")
    exists = True
except ResponseError:
    exists = False

# After
exists = client.exists("llama3.2")  # Returns True/False
```

## Changes

- Added `exists()` method to `Client` class
- Returns `True` if model is available locally, `False` otherwise
- Handles `ResponseError` internally
- Added tests for the new method

## Testing

- ✅ Test with existing model returns `True`
- ✅ Test with non-existing model returns `False`
- ✅ All existing tests pass

### DIFF
--- a/ollama/__init__.py
+++ b/ollama/__init__.py
@@ -52,6 +52,7 @@ push = _client.push
 create = _client.create
 delete = _client.delete
 list = _client.list
+exists = _client.exists
 copy = _client.copy
 show = _client.show
 ps = _client.ps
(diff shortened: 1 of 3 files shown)


ParthSareen:
The SDK is mostly a mirror for the API. You can just handle the exception with `except as...` and then do the appropriate behavior in your application.

PR #654, opened 2026-04-29
feat(types): add Audio class and audio field to Message for multimodal models

### LINKED ISSUE #650: Add dedicated 'audio' key for multimodal models
The Problem:
Currently, to pass audio data to multimodal models like gemma4:e2b, people are forced to use the images key.

Why this is an issue:
- Naming Confusion: Currently, audio data often has to be passed into a field called images. This is confusing because audio files are not images.
- Future Multimodal Support: For models that support both image and audio simultaneously, a single images bucket creates ambiguity.

Suggested Solution:
Add dedicated 'audio' key.

Ghraven:
## Summary

Adds a dedicated `Audio` class and `audio` field to `Message`, mirroring the existing `Image` pattern.

Closes #650

## Motivation

Currently, audio data must be passed via the `images` key, which is confusing and blocks future models that support both images *and* audio simultaneously. This PR adds a first-class `audio` field so callers can pass audio data cleanly:

```python
ollama.chat(
    model="gemma4:e2b",
    messages=[{
        "role": "user",
        "content": "Transcribe this",
        "audio": ["recording.wav"],   # clear, not crammed into images
    }]
)
```

## Changes

**`ollama/_types.py`**

- Added `Audio(BaseModel)` class after `Image`, with identical serialisation logic:
  - `Path` / `bytes` → base64-encodes the data
  - `str` path that exists on disk → base64-encodes the file
  - `str` with a known audio extension that doesn't exist → raises `ValueError` with a clear message
  - `str` that looks like existing base64 → passes through
  - Unknown string → raises `ValueError`
  - Supported extension check covers: `mp3`, `mp4`, `mpeg`, `mpga`, `m4a`, `ogg`, `wav`, `webm`

- Added `audio: Optional[Sequence[Audio]] = None` field to `Message` (after `images`), with the same docstring style as `images`.

## Compatibility

- No breaking changes — `audio` is optional and defaults to `None`
- The serialisation behaviour is consistent with `Image`, so the wire format is already what the Ollama server expects (raw base64)

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -186,6 +186,34 @@ class Image(BaseModel):
         raise ValueError('Invalid image data, expected base64 string or path to image file') from Exception
 
 
+class Audio(BaseModel):
+  value: Union[str, bytes, Path]
+
+  @model_serializer
+  def serialize_model(self):
+    if isinstance(self.value, (Path, bytes)):
+      return b64encode(self.value.read_bytes() if isinstance(self.value, Path) else self.value).decode()
+
+    if isinstance(self.value, str):
+      try:
+        if Path(self.value).exists():
+          return b64encode(Path(self.value).read_bytes()).decode()
+      except Exception:
+        # Long base64 string can't be wrapped in Path, so try to treat as base64 string
+        pass
+
+      # String might be a file path, but might not exist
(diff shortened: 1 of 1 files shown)


Ghraven:
Hi, just checking in! If multimodal audio support isn't on the roadmap yet or you'd prefer to handle it differently, totally understand — just let me know and I'll close it out. Otherwise happy to revise based on any feedback. Thanks for your time!

PR #671, opened 2026-05-22
feat: add tool_choice parameter to chat() (closes #663)

### LINKED ISSUE #663: Add tool_choice parameter to chat() (mirror server-side ollama#11171)
## Observation

The `chat()` function and underlying `ChatRequest` model in `ollama-python` do not expose a `tool_choice` parameter, even though several upstream endpoints already handle it:

- **OpenAI-compat endpoint** in `ollama/ollama` server (`openai/responses.go`): `ToolChoice` is parsed off the request, with TODO comments noting partial support.
- **Anthropic-compat endpoint** (`anthropic/trace.go` + `anthropic/anthropic.go`): `ToolChoice` is wired through.
- **Native `/api/chat`**: `tool_choice` is not a request field today; tracked upstream at ollama/ollama#11171 ("Add support for `tool_choice` set to `any` to require a model to use a tool").

### Empirical behavior today

```python
import ollama

# Test 1: tool_choice as kwarg → hard rejection
ollama.chat(
    model="llama3.1:8b",
    messages=[{"role": "user", "content": "Hello"}],
    tool_choice="auto",
)
# TypeError: Client.chat() got an unexpected keyword argument 'tool_choice'

# Test 2: tool_choice via options dict → silently accepted but not wired
# (options is passed through as model parameters; tool_choice has no effect there)
ollama.chat(
    model="llama3.1:8b",
    messages=[{"role": "user", "content": "Hello"}],
    options={"tool_choice": "auto"},
)
# OK, no error, but tool_choice is not interpreted by the server
```

`ChatRequest.model_json_schema()` confirms there is no `tool_choice` field. Current fields: `model`, `stream`, `options`, `format`, `keep_alive`, `messages`, `tools`, `think`, `logprobs`, `top_logprobs`.

### Why it matters

Practitioners porting code from OpenAI's or Anthropic's SDK to ollama-python expect a `tool_choice` parameter (to force tool use, restrict to a specific tool, or disable tools). Today the only available control is omitting `tools` entirely, which is coarse.

### Concrete suggestion

When the native `/api/chat` server endpoint adds `tool_choice` (per ollama/ollama#11171), expose the same parameter on `ollama.chat()` and `ChatRequest`:

```python
# Standard cross-vendor values
tool_choice: Literal["auto", "required", "none"] | dict | None = None
```

Until then, this issue serves as a tracking item for the Python client mirror so it lands in the same release window as the server-side feature.

### Workarounds

- For "must use a tool" semantics, append a directive to the system prompt and use `temperature=0.0` (verified to produce reliable tool calls on Llama 3.1 8B).
- For "must not use a tool", omit the `tools` argument entirely.

Happy to send a PR once the server-side direction is settled.

shivanireddyk:
Closes #663

Adds `tool_choice` support to `chat()` (sync and async), mirroring the OpenAI API and the server's existing partial support.

## Changes
- New `ToolChoiceFunction` class and `ToolChoice` type alias in `_types.py`
- Both exported from top-level `ollama` package
- `tool_choice` field added to `ChatRequest`
- All 6 `chat()` overloads updated (3 sync + 3 async) with typed parameter

## Usage
```python
from ollama import ToolChoiceFunction

# Forcing a specific tool
ollama.chat(
  model='llama3.2',
  tools=[get_weather],
  tool_choice=ToolChoiceFunction(
    function=ToolChoiceFunction.Function(name='get_weather')
  ),
  messages=[{'role': 'user', 'content': 'What is the weather in Paris?'}],
)

# It requires the model to call *some* tool
ollama.chat(model='llama3.2', tools=[...], tool_choice='required', messages=[...])
```


FF8:
I tested this PR against Ollama 0.32.5 with ollama-python `0.1.dev291+ge114845a0`. It looks like the client-side `tool_choice` parameter is accepted, but the native server does not appear to enforce it yet: `required` did not force a tool call, while `none` did not prevent one.

PR #677, opened 2026-06-04
fix: use delete=False for NamedTemporaryFile in tests on Windows

### LINKED ISSUE #676: Tests fail on Windows due to NamedTemporaryFile locking behavior
## Description
Several tests fail on Windows because `tempfile.NamedTemporaryFile()` 
locks the file while open, preventing other processes from reading it.
This works fine on Mac/Linux but not Windows.

## Failing Tests
- test_client_generate_images
- test_client_create_blob
- test_client_create_blob_exists
- test_async_client_generate_images
- test_async_client_create_blob
- test_async_client_create_blob_exists
- test_image_serialization_path
- test_image_serialization_string_path

## Environment
- OS: Windows
- Python: 3.14.3

## Error
PermissionError: [Errno 13] Permission denied: 'C:\Users\...\Temp\tmp...'

## Proposed Fix
Add `delete=False` to `NamedTemporaryFile()` calls in the affected tests.

athifmuhammad1001-svg:
Fixes #676  tests failing on Windows due to NamedTemporaryFile locking.

On Windows, NamedTemporaryFile locks the file while open, causing
PermissionError when the code tries to read it. Adding delete=False
resolves this.

Fixes: 8 failing tests on Windows
- test_client_generate_images
- test_client_create_blob
- test_client_create_blob_exists
- test_async_client_generate_images
- test_async_client_create_blob
- test_async_client_create_blob_exists
- test_image_serialization_path
- test_image_serialization_string_path

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -466,6 +466,11 @@ def test_client_generate_images(httpserver: HTTPServer):
     response = client.generate('dummy', 'Why is the sky blue?', images=[temp.name])
     assert response['model'] == 'dummy'
     assert response['response'] == 'Because it is.'
+  
+  try:
+    os.remove(temp.name)
+  except FileNotFoundError:
+    pass
 
 
 def test_client_generate_format_json(httpserver: HTTPServer):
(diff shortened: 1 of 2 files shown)


aglpy:
This fix solves the Windows issue, but since `delete=False` is used, the temp image files won’t be deleted automatically and may be left on disk.

Maybe it would be better to wrap this in a `try`/`finally` block and remove the file manually after it has been used.

Don’t you think?

athifmuhammad1001-svg:
> This fix solves the Windows issue, but since `delete=False` is used, the temp image files won’t be deleted automatically and may be left on disk.
> 
> Maybe it would be better to wrap this in a `try`/`finally` block and remove the file manually after it has been used.
> 
> Don’t you think?

Thanks for the insight! You were right, the temp files would remain on the disk when `delete=False` is used. I used `os.remove()` to explicitly remove those temp files using the `try`/`except` block in case of `FileNotFoundError`. I verified that the temp files do not remain on the disk after the changes were made. Pushed the fix.

PR #690, opened 2026-06-24
fix: preserve non-object response error bodies

lin-hongkuan:
### Summary
- Only read the `error` field from JSON object error responses.
- Preserve non-object JSON error bodies instead of raising a constructor-time exception.

### Why
`ResponseError` currently assumes any decoded JSON response is a dictionary. JSON strings, arrays, or numbers are valid response bodies and should still produce a useful `ResponseError`.

### Testing
- `python -m pytest tests/test_type_serialization.py::test_response_error_uses_json_object_error_field tests/test_type_serialization.py::test_response_error_preserves_non_object_json_body -q`
- `git diff --check`

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -638,7 +638,9 @@ class ResponseError(Exception):
     # try to parse content as JSON and extract 'error'
     # fallback to raw content if JSON parsing fails
     with contextlib.suppress(json.JSONDecodeError):
-      error = json.loads(error).get('error', error)
+      parsed_error = json.loads(error)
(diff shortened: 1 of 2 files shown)


vaderollama (approved):
LGTM — guards against non-dict JSON error bodies in `ResponseError.__init__`. Previously `json.loads(error).get('error', error)` would crash if the response body was a JSON string, array, or number. The `isinstance(parsed_error, dict)` check is the right fix. Tests cover both dict and non-dict cases.

PR #691, opened 2026-06-24
fix: strip whitespace from host values

lin-hongkuan:
### Summary
- Strip leading and trailing whitespace before parsing explicit host values.
- Cover host strings with schemes, ports, and paths.

### Why
Values copied into `OLLAMA_HOST` or passed as a host argument can include surrounding whitespace. Without trimming, host parsing can produce an invalid base URL or raise while parsing the port.

### Testing
- `python -m pytest tests/test_client.py::test_parse_host_strips_whitespace tests/test_client.py::test_headers -q`
- `git diff --check`

### DIFF
--- a/ollama/_client.py
+++ b/ollama/_client.py
@@ -1404,7 +1404,7 @@ def _parse_host(host: Optional[str]) -> str:
   'http://[0001:002:003:0004::1]:56789/path'
   """
 
-  host, port = host or '', 11434
+  host, port = (host or '').strip(), 11434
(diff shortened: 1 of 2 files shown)


vaderollama (approved):
LGTM — strips whitespace from host values before parsing, preventing invalid base URLs from copy-pasted `OLLAMA_HOST` values with trailing spaces/newlines. Tests cover scheme, port, and path combinations with spaces, tabs, and newlines. Small and correct.

PR #694, opened 2026-07-07
fix(example): use aware web search timestamps

Ghraven:
## Summary
- replace `datetime.utcnow()` in the GPT-OSS web search helper example with `datetime.now(timezone.utc)`
- keep the existing `Page.fetched_at` type and behavior while making timestamps timezone-aware

## Problem
The example currently creates fetched-at timestamps with naive UTC datetimes. `datetime.utcnow()` is deprecated in modern Python, and the resulting values do not carry explicit timezone information.

## Before / After
Before, fetched pages stored naive UTC timestamps.

After, fetched pages store UTC-aware timestamps with `timezone.utc`.

## Verification
- `python -m py_compile examples/web_search_gpt_oss_helper.py`
- `git diff --check`

### DIFF
--- a/examples/web_search_gpt_oss_helper.py
+++ b/examples/web_search_gpt_oss_helper.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import re
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Protocol, Tuple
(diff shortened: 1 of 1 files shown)


vaderollama (approved):
LGTM — replaces deprecated `datetime.utcnow()` with `datetime.now(timezone.utc)` in the web search example. Pure quality fix, no behavior change for users. Low risk.

PR #698, opened 2026-07-09
fix(types): expose tool call id on Message.ToolCall

### LINKED ISSUE #667: id in ToolCall missing
The id field in the tool_call is missing. The raw response (hook) contains the id (see below, "id": "call_p7o2gz50").
But the ToolCall class doesn't have this field nor is it in model_extra.

```json
{
  "model": "...",
  "created_at": "2026-05-18T...",
  "message": {
    "role": "assistant",
    "content": " ",
    "tool_calls": [
      {
        "id": "call_p7o2gz50",
        "function": {
          "index": 0,
          "name": "test_function",
          "arguments": {
            "param1": "test1",
            "param2": 123
          }
        }
      }
    ]
  },
  "done": true,
  "done_reason": "stop",
  ...
}
```

I have to use a hook to extract the id from the raw response and keep it in a list of the same size as the response.message.tool_calls list.

I suggest the following patch:

```diff
diff --git a/ollama/_types.py b/ollama/_types.py
index 96529d6..fdc00d2 100644
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -334,6 +334,8 @@ class Message(SubscriptableBaseModel):
     """
     Model tool calls.
     """
+    id: Optional[str] = None
+    'ID of the tool call.'
 
     class Function(SubscriptableBaseModel):
       """
```

winklemad:
## What & why

The Ollama server returns an `id` on each tool call in the response, e.g. `{"id": "call_p7o2gz50", "function": {...}}`. But `Message.ToolCall` has no `id` field, so it's silently dropped — and it isn't in `model_extra` either — leaving no way to read the tool-call id from Python (#667). That id is needed to correlate a tool call with its result.

## Change

Add an optional `id: Optional[str] = None` to `Message.ToolCall`. It's optional/defaulted so responses that omit it (older servers) keep parsing unchanged.

## Testing

`tests/test_type_serialization.py`: the id round-trips from a raw response and serializes back via `model_dump()`, and defaults to `None` when absent. Full suite: 94 passed. `ruff check` + `ruff format` clean.

Fixes #667

### DIFF
--- a/ollama/_types.py
+++ b/ollama/_types.py
@@ -346,6 +346,9 @@ class Message(SubscriptableBaseModel):
       arguments: Mapping[str, Any]
       'Arguments of the function.'
 
+    id: Optional[str] = None
+    'Identifier of the tool call, when provided by the server.'
+
     function: Function
     'Function to be called.'
 
--- a/tests/test_type_serialization.py
+++ b/tests/test_type_serialization.py
@@ -4,7 +4,7 @@ from pathlib import Path
 
 import pytest
 
-from ollama._types import CreateRequest, Image
+from ollama._types import CreateRequest, Image, Message
 
 
 def test_image_serialization_bytes():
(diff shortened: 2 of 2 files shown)


vaderollama (approved):
LGTM — adds the missing `id` field to `Message.ToolCall` matching the Go API (`ToolCall.Id` in `api/types.go`). Backwards-compatible (optional, defaults to None), tests cover round-tripping and default. Resolves #667.

winklemad:
Gentle nudge — this has an approval and is a small, backwards-compatible fix (exposes the `id` field on `Message.ToolCall`). Happy to rebase if anything has drifted. Thanks!

PR #700, opened 2026-07-13
style: ignore pylyzer warnings in _client.py

rpreslar4765:


vaderollama (commented):
Holding off on this one. The project uses ruff for linting (configured in `pyproject.toml`), not pylyzer. Adding a `# pylyzer: ignore` pragma at the top of `_client.py` introduces noise for a tool the project doesn't use.

If there are actual pylyzer warnings you're hitting locally, the better approach would be either:
1. Fix the underlying code issues pylyzer is flagging, or
2. Open a discussion about adopting pylyzer as a project tool

Happy to revisit if there's more context on which warnings prompted this.

PR #739, opened 2026-09-15
types: warn and drop options the server no longer supports

### LINKED ISSUE #18480: typical_p is no longer supported (status code: 400) from python API
### What is the issue?

Using https://github.com/ollama/ollama-python/releases/tag/v0.6.2

Image

New error from the server.   I saw the PR https://github.com/ollama/ollama/pull/18448 about TP but I thought it was still allowed.

### Relevant log output

```shell

```

### OS

Windows

### GPU

Nvidia

### CPU

AMD

### Ollama version

v0.34.2-rc0

dhiltgen:
Detect obsolete options and warn if they are detected, then strip them from the request.  Unknown options still pass through.  Adds min_p and draft_num_predict.

Fixes ollama/ollama#18480


pdevine (approved):
Looks fine to me. You should double check with @ParthSareen

ollama/_types.py
```diff
@@ -101,41 +104,78 @@ def get(self, key: str, default: Any = None) -> Any:
...
+  draft_num_predict: Optional[int] = None
 
   # runtime options
   num_keep: Optional[int] = None
   seed: Optional[int] = None
   num_predict: Optional[int] = None
   top_k: Optional[int] = None
   top_p: Optional[float] = None
-  tfs_z: Optional[float] = None
-  typical_p: Optional[float] = None
```
ParthSareen, line 127: Could we keep these fields around and just leave them out of requests? Existing code that reads `options.typical_p` now crashes, even though passing it into `Options` still works with a warning. A test that checks reading the old fields would help here.

ollama/_types.py
```diff
@@ -101,41 +104,78 @@ def get(self, key: str, default: Any = None) -> Any:
...
+    'typical_p',
+    'use_mlock',
+    'vocab_only',
+  }
+)
+
+
+def _drop_unsupported_options(options: Any) -> Any:
+  if not isinstance(options, Mapping):
+    return options
```
ParthSareen, line 129: This still lets `typical_p` through when someone uses `Options().model_copy(update={'typical_p': 0.5})`. I checked the request body and it’s sent without a warning, so the server would still reject it. Could we strip these options before sending the request and add a test for this case?

jessegross:
I'm not sure if we really want the SDKs to modify the requests - it seems like that logic should belong to Ollama itself so it is consistent. For example, if we want to allow GGUFs to continue to support typical_p in API requests that would probably be the friendliest way of deprecation or just strip there if that's what we want. But the SDKs should mark the field as deprecated in their docs and/or warn since otherwise many people will never know.

